PR #24, opened 2018-08-04
Update dep to 0.5

dnephin:


dnephin:
Uses go modules now. No need to update dep

PR #37, opened 2018-12-13
Add a godownloader compatible install script

sagikazarmark:
https://github.com/goreleaser/godownloader


dnephin:
Thanks for the contribution!

I have looked at godownloader. Generating a 300 line bash script that is meant to be executed by downloading it and piping it directly into `bash` seems like a strange way to solve the problem. I would rather not promote it as a recommended way to install.

If I understand correctly this script or the `godownloader` tool can be used to download the `gotestsum` binary without this script being added to the git repo.

sagikazarmark:
> Generating a 300 line bash script that is meant to be executed by downloading 

Well., this is already what you do with [dep](https://github.com/gotestyourself/gotestsum/blob/master/.circleci/config.yml#L77) and this is how quite a few projects [already work](https://install.goreleaser.com/projects/): hugo, golangci, goreleaser to name a few.

The godownloader tool generates this script, but in order to use it, one has to either download it during the build or generate the script in advance and commit it into the repository. The latter is only recommended when a project does not provide an install script, but the recommendation is to add this script to the repository.

The reason why I think it's a good thing is that the user does not have to guess where your prebuilt binaries are, what their name is. Just download the official install script, it can guess the latest version if you don't know one, job done.

Here is what I do at the moment to download gotestsum:

```
	@mkdir -p bin
ifeq (${OS}, Darwin)
	curl -L https://github.com/gotestyourself/gotestsum/releases/download/v${GOTESTSUM_VERSION}/gotestsum_${GOTESTSUM_VERSION}_darwin_amd64.tar.gz | tar -zOxf - gotestsum > ./bin/gotestsum-${GOTESTSUM_VERSION} && chmod +x ./bin/gotestsum-${GOTESTSUM_VERSION}
endif
ifeq (${OS}, Linux)
	curl -L https://github.com/gotestyourself/gotestsum/releases/download/v${GOTESTSUM_VERSION}/gotestsum_${GOTESTSUM_VERSION}_linux_amd64.tar.gz | tar -zOxf - gotestsum > ./bin/gotestsum-${GOTESTSUM_VERSION} && chmod +x ./bin/gotestsum-${GOTESTSUM_VERSION}
endif
```

Here is golangci compared:

```
curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b ./bin/ v${GOLANGCI_VERSION}
```

The second one seems to be easier, simpler.

Ideally I would like to have some sort of "binary package manager", that can download prebuilt binaries and is portable between operating systems:

```bash
$ bin get gotestyourself/gotestsum
```

I started experimenting with one, but haven't got anywhere (yet).

So in the meantime, I do think that an install script placed in the repository is the easiest way to install prebuilt binaries.

sagikazarmark:
I've started to work on https://github.com/sagikazarmark/binbrew

PR #46, opened 2019-03-06
Add a --no-cache option for avoiding cached test results

lox:
At the minute there doesn't seem to be a way to avoid the underlying behaviour of `go test` caching test results if nothing has changed. 

This behaviour was introduced in [Go 1.10](https://golang.org/doc/go1.10#test): "The go test command now caches test results"

Apparently "The idiomatic way to bypass test caching is to use -count=1".

This PR adds `--no-cache` which sets `--count=1`. IMO it's clearer, but we could also just expose `--count`. 

Btw thanks for the great tool!

### DIFF
--- a/main.go
+++ b/main.go
@@ -73,6 +73,7 @@ Formats:
 	flags.StringVar(&opts.junitFile, "junitfile",
 		lookEnvWithDefault("GOTESTSUM_JUNITFILE", ""),
 		"write a JUnit XML file")
+	flags.BoolVar(&opts.noCache, "no-cache", false, "disable test caching")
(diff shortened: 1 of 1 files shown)


dnephin:
Hi @lox , Thanks for the PR!

It is possible to set any `go test` arg by specifying it after a `--` (https://github.com/gotestyourself/gotestsum#custom-go-test-command)

```
gotestsum -- -count=1 ./...
```
I use this quite a bit to set things like `-coverprofile=coverage.txt`, or to specify a different test path. 

I'd like avoid adding new flags for all the `go test` flags. Does this option work for you?

lox:
Yup, sure does. Thanks @dnephin!

Issue #51, opened 2019-04-01
Add a --version flag


glenjamin:
I was checking whether an install script was working correctly, and instinctively ran `gotestsum --version`, which I discovered didn't exist.

I was going to send in a PR rather than report an issue, but I had a look at the deployment script and it's using some stuff i'm not familiar with - so I wasn't sure where to start.

If this is OKed, I'm happy to do the actual work if someone points me in the right direction.

dnephin:
I think there are some examples of passing values to the `go build` here: https://goreleaser.com/build/, with more details here: https://goreleaser.com/templates/

Looks like we can do something like: `-X main.version={{.Version}}`

To receive that value there needs to be a `var version string` in main.

glenjamin:
Looks like `goreleaser` sets some build vars by default: https://goreleaser.com/environment/#using-the-main-version

dnephin:
Nice, that is convenient

PR #52, opened 2019-04-02
Add a --version flag that goreleaser will populate for us

glenjamin:
I wasn't sure whether there was any mileage in trying to write some tests for this.


glenjamin:
Resolves #51

PR #57, opened 2019-06-20
Identify and, filter or display coverage output

### LINKED ISSUE #56: Coverage report duplicated in standard-quiet format
Hi, 

First, thanks for this tool, that's pretty cool and the final summary is really nice. 

The only thing that I miss was coverage support when coverage is enabled in go test. Switching to the format `standard-quiet` solve that by printing the normal line, but in this case, the coverage report is duplicated:
```
gotestsum --format standard-quiet ./test-folder1/... ./test-folder2/... ./local-code/... -- -cover
coverage: 50.6% of statements
ok  	path/to/some/tests	0.018s	coverage: 50.6% of statements
```
I tried to get a really quick look at the code but couldn't find immediately where was the issue. 
Also, if it was possible to have coverage support in short format (just having the percentage at the end of the line if coverage is enabled), that would be magic!

Thanks in advance, 
Moe

dnephin:
Closes #56

Identify coverage output events with string comparison. There does not appear to be another way.

Add coverage output to the short format, and filter out the duplicate line from the standard-quiet output.


Moe-Kite:
Just tested it quickly and it seems to work perfectly! Both in short and standard-quiet format.

 Thanks a lot! (and thanks for the reactivity, that's quite impressive :D )

PR #63, opened 2019-08-24
always colorize output unless specifically requested not to

ebilling:
Signed-off-by: Eric Billingsley <eric@calculi.com>


ebilling:
I want to preserve color even when running through a pipe.  I could do it this way, or with a forced color flag...

PR #70, opened 2019-10-05
Junit XML name formats

### LINKED ISSUE #50: Ability to apply a transformation to the "testsuite" name in junit
If I run tests on a package `github.com/ijc/foo/...` using `-junitfile` then I get results with:
```
<testsuite tests="33" failures="0" time="0.000000" name="github.com/ijc/foo/bar">
...
<testsuite tests="33" failures="0" time="0.000000" name="github.com/ijc/foo/baz">
```
...so far so good.

However if I feed this into Jenkins it creates me a directory named "github" and within that there are two entries "com/ijc/foo/bar" and "com/ijc/foo/baz", this is not really a desirable presentation.

It would be useful if gotestsum could allow fiddling with the name somehow. My preference would be to allow specifying a (common) prefix to be dropped from all the suite names (e.g. I would likely choose to drop `github.com/ijc/foo/`). A variant of that would be a `-p«N»` option (cf `patch(1)`) to strip «n» path elements.

I'm not sure if other transformations might be useful/better e.g. `tr . -` or some sort of escaping. 

Just stripping a prefix has the benefit of being nice and simple I think.

### LINKED ISSUE #68: Bitbucket Pipelines Test reporting displaying tests incorrectly
I'm using this with the [Bitbucket Pipelines Test reporting](https://confluence.atlassian.com/bitbucket/test-reporting-in-pipelines-939708543.html) and found an issue that test results displayed incorrectly.

Following input:
```xml
    <testsuite tests="14" failures="2" time="1.161" name="github.com/my_org/path/to/pkg">
        <properties>
            <property name="go.version" value="go1.13"></property>
        </properties>
        <testcase classname="github.com/my_org/path/to/pkg" name="TestInMyPkg" time="0.000">
            <failure message="Failed" type="">Example Error Message</failure>
        </testcase>
    </testsuite>
```

results in following test representation in Bitbucket Pipelines:
Screen Shot 2019-09-25 at 4 28 38 PM

While input with the short package name in the `classname` property:
```xml
    <testsuite tests="14" failures="2" time="1.161" name="github.com/my_org/path/to/pkg">
        <properties>
            <property name="go.version" value="go1.13"></property>
        </properties>
        <testcase classname="pkg" name="TestInMyPkg" time="0.000">
            <failure message="Failed" type="">Example Error Message</failure>
        </testcase>
    </testsuite>
```

Results in the much better representation on the Bitbucket Pipeliens UI:
Screen Shot 2019-09-25 at 4 31 09 PM

dnephin:
Fixes #50 
Fixes #68 
Closes #69 

Add new command line flags for configuring the format of the junit-xml output.

The flag names are unfortunately quite long, but I think shorter names might not be sufficient in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,17 @@ This file can be used to integrate with CI systems.
 gotestsum --junitfile unit-tests.xml
 ```
 
+If the package names in the `testsuite.name` or `testcase.classname` fields do not
+work with your CI system these values can be customized using the
+`--junitfile-testsuite-name`, or `--junitfile-testcase-classname` flags. These flags
+accept the following values:
+
+* `short` - the base name of the package (the single term specified by the 
+  package statement).
+* `relative` - a package path relative to the root of the repository
+* `full` - the full package path (default)
+
+
 ### JSON file output
 
 When the `--jsonfile` flag or `GOTESTSUM_JSONFILE` environment variable are set
--- a/flags.go
+++ b/flags.go
@@ -2,9 +2,11 @@ package main
 
 import (
 	"encoding/csv"
+	"path"
 	"strings"
 
 	"github.com/pkg/errors"
+	"gotest.tools/gotestsum/internal/junitxml"
 	"gotest.tools/gotestsum/testjson"
 )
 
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -86,6 +86,11 @@ This file can be used to integrate with CI systems.
 gotestsum --junitfile unit-tests.xml
 ```
 
+If the package names in the `testsuite.name` or `testcase.classname` fields do not
+work well with your CI system the values can be modified using one of the
+flags: `--junitfile-testsuite-name-format`, `--junitfile-testcase-classname-format`.
```
Trane9991: I think possible values (and differences between them) need to be documented as well.
Basically the: `"full, relative, short"`

flags.go
```diff
@@ -47,3 +49,38 @@ func (s *noSummaryValue) String() string {
...
+	return errors.Errorf("invalid value: %v, must be one of: "+junitFieldFormatValues, val)
+}
+
+func (f *junitFieldFormatValue) Type() string {
+	return "field-format"
+}
+
+func (f *junitFieldFormatValue) String() string {
+	return "full"
+}
```
Trane9991: Is it ok, that in any case `full` is returned here?

dnephin: Yup, this is what gets printed as the default value. Since there is no way to change the default value, this should always be correct.

dnephin:
Thanks for the suggestion! I've improve the README by including the supported values and an explanation of each value

Trane9991:
@dnephin could you release a new binary version with this fix, please?

dnephin:
Yes, I am going to try and get one or 2 more changes in and I should release a new version within a week.

dnephin:
https://github.com/gotestyourself/gotestsum/releases/tag/v0.4.0

PR #73, opened 2019-10-13
Add short-with-failures format

### LINKED ISSUE #72: Configure verbosity of failing tests separately from passing tests
When running longer test suites, I don't want to wait for the suite to finish to see my failures. Currently the only way to get this behavior is to use a `-verbose` format, but then I see all of my passing tests, which makes it hard to see the (comparatively) smaller number of failures.

I would like to be able to specify whether to show passing tests verbosely independently of failing tests. Something like `gotestsum --format short --verbose-failures`.

dnephin:
Closes #72

Adds a new format, similar to `short`, that prints test output as soon as a test fails. Useful when a test suite runs for a long time and you want to see errors before the run ends.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ gotestsum --format short-verbose
 Supported formats:
  * `dots` - output one character per test.
  * `short` (default) - output a line for each test package.
+ * `short-with-failures` - like short, but prints test output as soon as a test
+   fails.
  * `standard-quiet` - the default `go test` format.
  * `short-verbose` - output a line for each test and package.
  * `standard-verbose` - the standard `go test -v` format.
--- a/README.md
+++ b/README.md
@@ -37,12 +37,11 @@ gotestsum --format short-verbose
 ```
 
 Supported formats:
- * `dots` - output one character per test.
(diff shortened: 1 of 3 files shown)


wendorf:
Works great! Thanks for doing this. The help text needs updating, though:

```
diff --git a/main.go b/main.go
index cd9b7c9..81cc97e 100644
--- a/main.go
+++ b/main.go
@@ -65,11 +65,12 @@ Flags:
 		flags.PrintDefaults()
 		fmt.Fprint(os.Stderr, `
 Formats:
-    dots              print a character for each test
-    short             print a line for each package
-    short-verbose     print a line for each test and package
-    standard-quiet    default go test format
-    standard-verbose  default go test -v format
+    dots                 print a character for each test
+    short                print a line for each package
+    short-verbose        print a line for each test and package
+    short-with-failures  print a line for each test and package
+    standard-quiet       default go test format
+    standard-verbose     default go test -v format
 `)
 	}
 	flags.BoolVar(&opts.debug, "debug", false, "enabled debug")
```

dnephin:
Good catch, thanks!

PR #84, opened 2020-01-05
Documentation for executing test binary

### LINKED ISSUE #77: how to use this in e2e tests binaries?
I am interested to add this to minikube e2e tests, but I am not sure how I would use it, we build e2e binaries for the the tests and run it on different platforms. could I still use this tool?

tiny-dancer:
Closes #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,8 @@ output.
 - [Summary](#summary)
 - [JUnit XML](#junit-xml)
 - [JSON file](#json-file-output)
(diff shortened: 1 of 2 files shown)


dnephin:
Thank you for the PR! This is a great addition to the documentation.

I pushed another commit which makes some typographical changes to keep the readme consistent, and I moved the bit about "running without Go" into a separate file, as I expect that use case is a little more rare. It's linked from the same place in the README, so should be easy to find.

Also squashed a few commits.

PR #90, opened 2020-02-22
Add slowest tests to summary

dnephin:
Add a flag for setting the slow test threshold, defaults to 100ms.

Add a count of the number of slow tests to the summary line, and list the 5 slowest tests above the threshold.


dnephin:
Now that `gotestsum tool slowest` exists there is probably no need to add slowest to the summary.

Issue #66, opened 2019-09-16
test2json incorrectly attributes output to a subtest, resulting in missing test failure output


dnephin:
Upstream issue: https://github.com/golang/go/issues/29755

Symptoms: `gotestsum` reports that a test failed, but does not show any failure output in the summary.

Workarounds: Use the `standard-verbose` format.

Possibly solution: If  the bug is not fixed upstream then one solution would be to print the output for the entire top level test, instead of just the output for the subtest.

dnephin:
https://github.com/golang/go/pull/34419
https://go-review.googlesource.com/c/go/+/196617

dnephin:
The fix for this was included in go1.15, so I think it's probably time to close this issue.

edaniels:
Without using the workaround, I still see this issue with standard arguments. Including --debug has it sometimes show up.

dnephin:
Thanks for the report! Would you mind opening a new issue with info about the Go version used, and if possible steps to reproduce?

From the linked pull request it sounds like maybe it has to do with `go test -race` output?

edaniels:
Filed https://github.com/gotestyourself/gotestsum/issues/310

Issue #64, opened 2019-08-30
Race Detection not mentioned in summary, giving confusing results


howardjohn:
When `-race` is used an a race is detected, I get

```
=== FAIL: galley/pkg/config/source/kube/fs TestAddUpdateDelete (0.04s)

```
(no info other than FAIL)

But when a test failures I get a message like

```
=== FAIL: galley/pkg/config/source/kube/fs TestAddUpdateDelete (0.00s)
    source_test.go:141: my test failed
```

It may be useful to mention a data race was detected or something here?

dnephin:
Thanks for the repo! This is something that I would like to fix/improve.

dnephin:
Do you have any way to reproduce the issue or examples of how `go test -v` displays the race failures? I'm not sure where  the race is reported in the test2json output.

howardjohn:
Here is one example I have off hand: https://storage.googleapis.com/istio-prow/logs/integ-race-native-tests_istio_postsubmit/14/build-log.txt (search DATA RACE)

I can make a more minimal reproduction in a bit

dnephin:
I believe those failures are coming from a subtest, is that correct?

The underlying issue may be an a bug in `test2json` output. I put some notes into: #66

dnephin:
I was able to reproduce a data race failure, and I saw the output using `short-verbose` format, so I believe that confirms the issue is #66.

dnephin:
[gotestsum v0.4.2](https://github.com/gotestyourself/gotestsum/releases/tag/v0.4.2) includes a work around for the bug in `test2json`. I'm still hoping my patch to fix the problem is accepted in Go before 1.15, but until then hopefully the workaround addresses the problem.

I'll keep #66 open to track the issue.

PR #102, opened 2020-04-19
mitigate test2json bug causing missing test output

dnephin:
See golang/go#29755, which may be fixed by https://github.com/golang/go/pull/34419 if that is merged. Even if that fix makes it into go1.15 there will be plenty of projects still running earlier versions of Go, so a workaround may be useful for another year or more.

Related to #64, #66

If a test fails, and there is no test output (other than framing), then print all the test output for the entire test case.

This won't fix the issue if the test which failed had some output, but well behaved tests generally shouldn't non-failure message output, so hopefully this case is rare.

### DIFF
--- a/internal/junitxml/report.go
+++ b/internal/junitxml/report.go
@@ -155,14 +155,16 @@ func packageTestCases(pkg *testjson.Package, formatClassname FormatFunc) []JUnit
 		jtc := newJUnitTestCase(tc, formatClassname)
 		jtc.Failure = &JUnitFailure{
 			Message:  "Failed",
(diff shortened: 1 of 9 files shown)


dnephin:
The problem with this workaround is that any failed subtest will cause the root test to print all the output for all subtests in the summary.

I think it's going to require another check for the root test case to only add the extra files if none of the subtests have failed.

dnephin:
I think I have resolved the above issue by including the changes from #104 and adding a `subTestFailed` field to `TestCase`.

PR #112, opened 2020-05-16
Windows: use golang.org/x/sys to detect Terminal

thaJeztah:


thaJeztah:
@dnephin ptal

(looks like it's still used by other dependencies though)

dnephin:
Ya, looks to be still used by `github.com/fatih/color`

dnephin:
Thanks! I rebased to remove the merge config, and updated another dependency while I was at it

thaJeztah:
Thanks! Have one more thing that looks like it can be replaced; let me push that

PR #113, opened 2020-05-16
do: use GOLANG_VERSION arg

dnephin:
pass the arg, and use it in the iidfile cache filename so that using a different value doesn't keep using the same cached image id.

### DIFF
--- a/do
+++ b/do
@@ -20,24 +20,31 @@ go-mod-tidy() {
     git diff --stat --exit-code go.mod go.sum
 }
 
+help[shell]='Run a shell in a golang docker container.
+
+Env vars:
+
+GOLANG_VERSION - the docker image tag used to build the image.
+'
 shell() {
-    local tag; tag="$(_docker-build-dev)"
+    local image; image="$(_docker-build-dev)"
     docker run \
         --tty --interactive --rm \
         -v "$PWD:/work" \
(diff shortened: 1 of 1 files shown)


do
```diff
@@ -20,36 +20,43 @@ go-mod-tidy() {
...
+    local image; image="$(_docker-build-dev)"
     docker run \
         --tty --interactive --rm \
         -v "$PWD:/work" \
         -v ~/.cache/go-build:/root/.cache/go-build \
         -v ~/go/pkg/mod:/go/pkg/mod \
         -w /work \
-        "$tag" bash
+        "$image" \
+        "${@-bash}"
```
thaJeztah, line 38: Wondering if `DOCKER_BUILDKIT=1` is an option, in that case, you could use the `docker build --output=..` option to get artifacts out of the build, instead of first building, then running the image

dnephin: Oh interesting!

In this case I don't think there will necessarily be any output. The goal is to run an interactive shell and potentially run multiple commands in it. 

If someone wants to try something against a different Go version it is a convenient way of getting a container running with that version.

thaJeztah: Ah, gotcha; didn't look closely at all use-cases, but the `--output` option is very interested for "building" artefacts (and can make things a lot more simple). Only issue would be if you want to do so with native Windows containers (windows doesn't yet support buildkit)

PR #114, opened 2020-05-16
update pkg/errors to v0.9.1

thaJeztah:
This version plays well with go1.13 and up "wrapped" errors, and makes sure that `errors.Unwrap()`, `errors.Is()` and `errors.As()` are able to get the wrapped error.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/jonboulle/clockwork v0.1.0
 	github.com/mattn/go-colorable v0.0.9 // indirect
 	github.com/mattn/go-isatty v0.0.3
-	github.com/pkg/errors v0.8.1
+	github.com/pkg/errors v0.9.1
 	github.com/spf13/pflag v1.0.3
 	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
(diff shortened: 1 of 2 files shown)


thaJeztah:
Once go < 1.13 is no longer supported by this library, `GetExitCode` could be rewritten to something like;

```go
// GetExitCode returns the ExitStatus of a process from the error returned by
// exec.Run(). If the exit status is not available an error is returned.
func GetExitCode(err error) (int, error) {
	var exitErr exec.ExitError
	if errors.As(err, &exitErr) {
		if procExit, ok := exitErr.Sys().(syscall.WaitStatus); ok {
			return procExit.ExitStatus(), nil
		}
	}
	return 0, errors.Wrap(err, "failed to get exit code")
}
```

dnephin:
Thanks! 

I think once go1.11 and go1.12 support is dropped I will probably remove the dependency and use `%w`.

thaJeztah:
> I think once go1.11 and go1.12 support is dropped I will probably remove the dependency and use %w.

Be careful doing so, because if "downstream" users of this package (not sure if that's something in practice) didn't switch to replace `errors.Cause()` to `errors.Is()`, they won't get the underlying error

PR #115, opened 2020-05-16
Use golang.org/x/sys/windows.GetConsoleScreenBufferInfo

thaJeztah:
Looks like this is now provided by golang.org/x/sys as well. The other ones not yet (afaics)

Perhaps  @tklauser knows if that's within scope of golang.org/x/sys, or if we need to continue implementing it here

(note; I don't have a Windows machine running, so have not tested it)

### DIFF
--- a/internal/dotwriter/writer_windows.go
+++ b/internal/dotwriter/writer_windows.go
@@ -15,7 +15,6 @@ import (
 var kernel32 = syscall.NewLazyDLL("kernel32.dll")
 
 var (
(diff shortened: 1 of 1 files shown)


dnephin:
Thanks!

tklauser:
> Looks like this is now provided by golang.org/x/sys as well. The other ones not yet (afaics)
> 
> Perhaps @tklauser knows if that's within scope of golang.org/x/sys, or if we need to continue implementing it here

I think it would be perfectly acceptable to add the other functions to x/sys/windows as well. Please feel free to send a PR/CL if needed.

PR #126, opened 2020-06-06
Rerun failed tests

dnephin:
Adds a new `--rerun-fails[=n]` flag. When the flag is used `gotestsum` will rerun any failed tests up to `n` times until each one passes at least once.

The new feature can be used to mitigate the pain of flaky tests in CI. To avoid unnecessarily rerunning tests when there is a bug or problem with the test environment, if there are more than 10 failed tests, rerun will be skipped. The `--rerun-fails-max-failures` flag can be used to change that threshold.

Using the `--rerun-fails` flag along with `--raw-command` or with `go test` args may require special handling, which will need to be documented in #117

### DIFF
--- a/rerunfails.go
+++ b/rerunfails.go
@@ -1,0 +1,113 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"gotest.tools/gotestsum/testjson"
+)
+
+type rerunOpts struct {
+	runFlag string
+	pkg     string
+}
+
+func (o rerunOpts) Args() []string {
+	var result []string
+	if o.runFlag != "" {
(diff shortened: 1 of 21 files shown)


rerunfails.go
```diff
@@ -0,0 +1,112 @@
...
+func goTestRunFlagFromTestCases(tcs []string) string {
+	buf := new(strings.Builder)
+	buf.WriteString("-run=")
+	for i, tc := range tcs {
+		if i != 0 {
+			buf.WriteString("|")
+		}
+		buf.WriteString(tc)
+	}
+	return buf.String()
```
dnephin, line 111: I think this needs a `^...$` to ensure that it doesn't prefix-match other tests.

The existing E2E test can be augmented to show that working.

PR #153, opened 2020-10-12
Add --watch flag for running tests when files change.

dnephin:
This just happens to work super well with #147. Sending `ctrl-c` while tests are running will cancel the run. Sending `ctrl-c` when no tests are running will exit.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,10 +8,12 @@ import (
 	"os/exec"
 	"os/signal"
 	"strings"
+	"time"
 
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
+	"gotest.tools/gotestsum/internal/filewatcher"
 	"gotest.tools/gotestsum/log"
 	"gotest.tools/gotestsum/testjson"
 )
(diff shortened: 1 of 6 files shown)


cmd/main.go
```diff
@@ -30,13 +32,32 @@ func Run(name string, args []string) error {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
+	defer cancel()
+
+	fn := func(pkg string) error {
+		opts := *opts
+		opts.packages = []string{pkg}
+		err := run(&opts)
+		if ExitCodeWithDefault(err) > 1 {
+			return err
+		}
```
dnephin, line 55: Build errors exit with 2. This should only error if it's not an `exitCoder` error.

internal/filewatcher/watch.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil
+	}
+
+	if time.Since(h.last) < floodThreshold {
+		log.Debugf("skipping event received less than %v after the previous", floodThreshold)
+		return nil
+	}
+
+	pkg := "./" + filepath.Dir(event.Name)
+	fmt.Printf("\nRunning tests for %v\n", pkg)
```
dnephin, line 173:
```suggestion
	fmt.Printf("\nRunning tests in %v\n", pkg)
```

PR #164, opened 2020-11-21
ci: update go version on windows

dnephin:


dnephin:
It's unfortunate that it takes 2+ minutes to install Go, but it's probably better the tests run against the latest release version instead of go1.12, which we'll probably remove from supported version once go1.16 is released.


The [github actions image includes go1.14](https://github.com/actions/virtual-environments/blob/releases/win19/20201021/images/win/Windows2019-Readme.md). If that image is kept more up to date, that could be another option.

PR #169, opened 2020-12-19
ci: Add go 1.16

dnephin:
Also remove go1.12, and add some docs for a `--watch` feature that was added recently.


dnephin:
I'm not sure why the github status is not being updated, but the CI run was green here: https://app.circleci.com/pipelines/github/gotestyourself/gotestsum/355/workflows/5fd760cc-32b1-47c1-9ce9-106d748c300d

PR #172, opened 2021-01-15
check err values for WriteString

docwhat:
In the name of being thorough...  check all errors.

### DIFF
--- a/cmd/watch.go
+++ b/cmd/watch.go
@@ -97,7 +97,9 @@ func delveInitFile(exec *testjson.Execution) (string, func(), error) {
 	for _, tc := range exec.Failed() {
 		fmt.Fprintf(buf, "break %s\n", tc.Test.Name())
 	}
-	buf.WriteString("continue\n")
+	if _, err := buf.WriteString("continue\n"); err != nil {
+		panic(err)
+	}
 	if err := buf.Flush(); err != nil {
 		remove()
 		return "", nil, err
--- a/log/log.go
+++ b/log/log.go
@@ -20,6 +20,14 @@ var (
 	out   stringWriter = os.Stderr
 )
 
+func writeOrPanic(s string) int {
+	n, err := out.WriteString(s)
+	if err != nil {
+		panic(err)
+	}
+	return n
+}
+
 // TODO: replace with io.StringWriter once support for go1.11 is dropped.
 type stringWriter interface {
 	WriteString(s string) (n int, err error)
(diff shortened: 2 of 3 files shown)


cmd/watch.go
```diff
@@ -97,7 +97,9 @@ func delveInitFile(exec *testjson.Execution) (string, func(), error) {
 	for _, tc := range exec.Failed() {
 		fmt.Fprintf(buf, "break %s\n", tc.Test.Name())
 	}
-	buf.WriteString("continue\n")
+	if _, err := buf.WriteString("continue\n"); err != nil {
+		panic(err)
+	}
```
dnephin, line 102: This `buf` here is a `bufio.Writer` which has a really cool feature: https://golang.org/pkg/bufio/#Writer

>  If an error occurs writing to a Writer, no more data will be accepted and all subsequent writes, and Flush, will return the error

What this allows us to do is perform multiple writes (or in this case `WriteString`) and ignore the error. After this line we `Flush()` the buffer, and handle the error.  So even though we aren't handling the error on this line, we do correctly handle any error this write may have performed in the call to Flush below.

The linter doesn't complain, but our `bufio.Writer` is actually capture errors for all the `fmt.Fprintf` calls in the loop above as well. The linter defaults to ignoring errors from `Fprintf` (for some reason, I'm not exactly sure why).

I think this is a really nice pattern for error handling. I've seen it mentioned in an article by Rob Pike (although sadly I can't find that link right now), [this article from Dave Cheney](https://dave.cheney.net/2019/01/27/eliminate-error-handling-by-eliminating-errors) goes into a bit as well.

log/log.go
```diff
@@ -35,36 +43,36 @@ func Warnf(format string, args ...interface{}) {
 	if level < WarnLevel {
 		return
 	}
-	out.WriteString(color.YellowString("WARN "))
-	out.WriteString(fmt.Sprintf(format, args...))
-	out.WriteString("\n")
+	writeOrPanic(color.YellowString("WARN "))
+	writeOrPanic(fmt.Sprintf(format, args...))
+	writeOrPanic("\n")
```
dnephin, line 48: All the write calls in this file are for logs. Logs end up being a bit of a special case because if a log fails there ends up not really being any other good way to report the problem. The panic output is going to end up being sent to the same file handle as the logs, so if we can't write logs, then the user also won't see the panic output.

That leaves the question of: "if we fail to log, should the program continue to run or should it exit".  I guess there is an argument to be made either way. I think I'm still slightly on the side of "let the program continue until it hits an error that prevents it from performing its primary function.

dnephin (commented):
Thank you for the PR! I appreciate the desire to ignore fewer linter warnings.

I left more detailed comments below, but I think in this case ignoring the warning is safe. In the future, errors from other `WriteString` calls may need to be handled, but often I find that `WriteString` is used to write to buffers which can never error, so I think in general it's going to be ok to ignore this linter warning.

log/log.go
docwhat: I'm from the "if something goes wrong, leave a smoking crater" school of thought.

Article by me: [Smoking craters are good](https://docwhat.org/smoking-craters-are-good)

PR #177, opened 2021-02-17
remove dnephin/pflag replace

### LINKED ISSUE #176: Can't install with `go install` in 1.16 due to use of pflag fork
The use of a replace directive to use a fork of `pflag` causes issues:
https://github.com/gotestyourself/gotestsum/blob/6d0e3109cfd4f5fcce05a9120f10c5193ab55e42/go.mod#L3

```
$ cd $(mktemp -d)
/tmp/tmp.3upu9q0UQv$ go install gotest.tools/gotestsum@latest
go install gotest.tools/gotestsum@latest: gotest.tools/gotestsum@v1.6.1
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

I think the replace directive can be removed, however, as github tells me that the fork in question has been fully merged to upstream: https://github.com/spf13/pflag/compare/master...dnephin:master
> spf13:master is up to date with all commits from dnephin:master

adamko147:
Removed the `dnephin/pflag` replace and run `go mod tidy`
Closes #176


adamko147:
@dnephin can you please check? also, no sure why the build on CI fails, the error is about missing GCC.
Thanks, I would really love to have this package installable using `go install` with go1.16

dnephin:
The build failures should be unrelated. The windows once seems to be that the Chocolatey package for go1.16 is not installing into PATH, or is otherwise somehow different from previous versions. I'll have to look into that.

The arm64_darwin I think was a new addition, I'm not sure why the failure wasn't detected when it was added. I'll look into that as well.

I'll comment on the issue about specifics of this change.

dnephin:
The `arm64_darwin` failure looks like it was related to https://github.com/goreleaser/goreleaser/pull/2069. They added support, and some configs would automatically start trying to build it.  I ran another build for #178 and that is no longer failing.

However the windows build is still failing.

PR #181, opened 2021-02-21
Added support for 'arm' binary release

### LINKED ISSUE #180: Add binary release for 32-bit ARM?
Thanks for the great tool!

I'm trying to use `gotestsum` in my CI/CD pipeline which uses GH Actions + couple self-hosted runners on Raspberry PI (for ARM+ARM64). Installing via `go get` on 32-bit ARM CI/CD runner is very slow (1 minute 20 seconds on Raspberry PI 4).
I'd love to be able to download pre-built binary, but it's only available for `arm64` and not `arm`.

jkowalski:
Fixes #180

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,6 +15,7 @@ builds:
     goarch:
       - amd64
       - arm64
+      - arm
     env: [CGO_ENABLED=0]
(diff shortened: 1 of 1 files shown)


dnephin:
Thanks!

PR #183, opened 2021-03-14
replace all usages of os/exec with golang.org/x/sys/execabs

thaJeztah:
follow-up to https://github.com/gotestyourself/gotestsum/pull/182, only the last commit is new

Following the changes in Go, and golang.org/x/tools themselves, this change
ensures that packages using exec.LookPath or exec.Command to find or run
binaries do not accidentally run programs from the current directory when
they mean to run programs from the system PATH instead.

### DIFF
--- a/cmd/handler.go
+++ b/cmd/handler.go
@@ -4,9 +4,9 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
 
 	"github.com/pkg/errors"
+	exec "golang.org/x/sys/execabs"
 	"gotest.tools/gotestsum/internal/junitxml"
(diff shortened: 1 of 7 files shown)


dnephin (approved):
LGTM

thaJeztah:
rebased to get rid of the extra commits, and moved out of draft

dnephin:
Thanks for working on this! Sorry we never got it merged. It looks like these fixes are now in the stdlib: https://golang.org/doc/go1.19#os-exec-path

I've updated the released builds to use go1.19, so I think we can probably stick to the stdlib now.

PR #184, opened 2021-03-24
move JUnitTest* structs to public package

nakatamixi:
Hi!

I love to use `gotestsum`.

This PR move `JUnitTest*` structs to public.
Users can parse `junitfile` for motivation such as exporting test metrics.

How do you think about publicization these structs?
Please review this.


dnephin:
Hello, and thank you for your interest in gotestsum! 

Unfortunately I would like to keep these types in an internal package. JUnit XML is difficult because there is no official spec. Every platform that uses it has some slightly different behaviour. I am not even sure if these types are correct for all use cases. I think it might be better to instead copy these types into whatever package needs to use them.

nakatamixi:
OK.
I understand it.
Thanks to reply!

Issue #141, opened 2020-08-09
Passing tests reported as failures in summary


mattmoor:
We are using `gotestsum` to run our testing for Knative, and digging into some recent "failures", I can't really make sense of why `gotestsum` is reporting them as failed.

The failure log I have been digging through can be downloaded from [here](https://console.cloud.google.com/storage/browser/knative-prow/logs/ci-knative-serving-contour-latest/1292502104706912256?prefix=), the file you want is "build-log.txt" (it is large).

In this test run I see the following summary:

```
=== Skipped
=== SKIP: test/e2e TestProbeWhitelist (0.00s)
    probe_whitelist_test.go:38: RequestAuthentication dropped the regex support. We need to rewrite this test. See https://github.com/istio/istio/issues/16585


=== Failed
=== FAIL: test/conformance/api/v1alpha1 TestContainerExitingMsg/http (unknown)

=== FAIL: test/conformance/api/v1alpha1 TestContainerExitingMsg/tcp (unknown)
```

However, if I scan back through the test logs, I see that all of the flavors of `TestContainerExitingMsg` have passed, here's one cluster:

```
--- PASS: TestContainerExitingMsg (0.00s)
    --- PASS: TestContainerExitingMsg/http (40.29s)
    --- PASS: TestContainerExitingMsg/tcp (172.49s)
```

I've been trying to make sense of the pieces of information here, but basically all I've managed so far is that `(unknown)` seems to tell us that the elapsed time is unknown, which seems to point to a TestCase terminated by the `end()` method [here](https://github.com/gotestyourself/gotestsum/blob/e0763274cff147121c0c8f53e84b751843f2219c/testjson/execution.go#L220).

Any pointers would be appreciated. 🙏

dnephin:
`unknown` indicates that `gotestsum` thinks there was a test that was missing a PASS/FAIL event when the test run completed. It was added in #104 to mitigate https://github.com/golang/go/issues/38382

But I also don't see an incomplete test here. It seems like `TestContainerExitingMsg` was run 3 times, and passed 3 times. 

If you can add a `--jsonfile` flag, that will save the source `test2json` output that `gotestsum` receives. We might get a better idea from that log, or at least find a way to reproduce the issue.

Are you running `gotestsum` from master? Do you know when this first started happening? Are you running with `-count=3` or is something else causing the tests to run multiple times?

mattmoor:
[...]

PR #188, opened 2021-04-01
Handle missing "pass" event for subtests

aaronlehmann:
Sometimes no "pass" event is present for subtests in the JSON output,
even though the output indicates the subtests passed (with "--- PASS:")
and the parent test had a pass event. In these cases, gotestsum can
output spurious failure messages for the subtests. Handle this by
treating a pass event for a parent test as implying that the subtests
also passed.

### DIFF
--- a/testjson/execution.go
+++ b/testjson/execution.go
@@ -72,10 +72,10 @@ type Package struct {
 	Skipped []TestCase
 	Passed  []TestCase
 
-	// mapping of root TestCase ID to all sub test IDs. Used to mitigate
-	// github.com/golang/go/issues/29755.
-	// In the future when that bug is fixed this mapping can likely be removed.
-	subTests map[int][]int
+	// mapping of root TestCase ID to most recent subtest TestCases. Used to
+	// mitigate github.com/golang/go/issues/29755, and also to infer passage of
+	// subtests from passage of the parent test.
+	subTests map[int][]TestCase
 
 	// output printed by test cases, indexed by TestCase.ID. Package output is
 	// saved with key 0.
@@ -156,7 +156,7 @@ func (p *Package) OutputLines(tc TestCase) []string {
 	result := make([]string, 0, len(lines)+1)
 	result = append(result, lines...)
 	for _, sub := range p.subTests[tc.ID] {
-		result = append(result, p.output[sub]...)
+		result = append(result, p.output[sub.ID]...)
 	}
 	return result
 }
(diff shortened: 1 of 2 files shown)


testjson/execution.go
```diff
@@ -344,6 +344,10 @@ func (p *Package) addTestEvent(event TestEvent) {
 	}
 
 	// the event.Action must be one of the three test end events
+	p.addEndTestEvent(event, tc)
+}
+
+func (p *Package) addEndTestEvent(event TestEvent, tc TestCase) {
```
aaronlehmann, line 350: `gocyclo` forced this change on me. I'm personally not a fan of `gocyclo` and how it can force small fixes to be bundled with refactors, but hey, it's not my codebase.

aaronlehmann:
@coryb pointed out to me that the root cause actually looks like `go test` misattributing certain output lines. For example, we've seen `--- PASS` lines with the wrong `Test` JSON key.

So this PR might help with the symptom, but since there appears to be a deeper issue with bad data coming from the JSON, I'm not sure it makes sense to move forward with this.

dnephin:
Hi @aaronlehmann , thank you for the PR! Yes this issue sounds familiar. I think #141 has some investigation and a link to the upstream Go issue (https://github.com/golang/go/issues/38063). From that discussion it seems like the issue is more common with parallel tests.

As you found in the code there is already some logic to handle other `go test` or `test2json` bugs, so I'm not against trying to fix it in `gotestsum`.  I'll give this PR a read over the weekend and see if we can get it merged.

Which version of Go are you using when you see these bugs? I believe the misattributing lines was a serious regression in 1.14.x , and was eventually fixed in 1.15.x. There are [a few closed issues tagged with `test2json-bug`](https://github.com/gotestyourself/gotestsum/issues?q=label%3Atest2json-bug+is%3Aclosed) that document those problems.

aaronlehmann:
I'm seeing this with Go 1.14.7. I suspect my problem is related to tests writing directly to stdout/stderr. That sounds like unsupported behavior. Unfortunately it will be a bit painful to fix.

I found this upstream issue that looks related: https://github.com/golang/go/issues/38050

It was ultimately fixed by tweaking the test output, rather than adjusting the parser: https://github.com/golang/go/commit/2ba9d45f9cc033bfc584701f458a032aa04f89c5

...so it's not a huge surprise that tests writing directly to stdout/stderr might cause the parser to get confused in a similar way.

aaronlehmann:
I believe the specific issue I'm hitting is https://github.com/golang/go/issues/40771. It sounds like a fix is included in Go 1.14.10 (and 1.15.3, and 1.16). So upgrading Go will likely fix the issue for me. I'll close this since it looks like upgrading Go to get the bug fix is a better path forward.

PR #191, opened 2021-04-17
Mitigate go test bug with missing subtest pass events 

### LINKED ISSUE #141: Passing tests reported as failures in summary

dnephin:
Inspired by the changes in #188
Fixes #141

Previously all missing events were reported. This commit omits events for any missing subtests pass events, so that subtests that passed do not show as failed.

The upstream bug was fixed in go1.16, so this mitigation probably does not need to stick around for long.

@aaronlehmann I added you as a co-author to this commit since this was heavily inspired by your change. I hope that is ok.

### DIFF
--- a/testjson/execution.go
+++ b/testjson/execution.go
@@ -73,8 +73,8 @@ type Package struct {
 	Passed  []TestCase
 
 	// mapping of root TestCase ID to all sub test IDs. Used to mitigate
-	// github.com/golang/go/issues/29755.
-	// In the future when that bug is fixed this mapping can likely be removed.
+	// github.com/golang/go/issues/29755, and github.com/golang/go/issues/40771.
+	// In the future when those bug are fixed this mapping can likely be removed.
 	subTests map[int][]int
 
 	// output printed by test cases, indexed by TestCase.ID. Package output is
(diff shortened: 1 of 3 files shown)


aaronlehmann (approved):
LGTM

PR #197, opened 2021-05-28
enable running all tests when watching by pressing 'a'

dprotaso:


dprotaso:
Bit torn on the lint error - fixing that means this diff comes much larger since it requires refactoring that method

> internal/filewatcher/watch.go:24:1: cyclomatic complexity 13 of func `Watch` is high (> 12) (gocyclo)
func Watch(dirs []string, run func(opts RunOptions) error) error {

Few options - so let me know which is preferable 
1) Refactor the method in this PR
2) Ignore the lint 
3) Bump the threshold to cyclomatic complexity to 13 so it can be fixed in a separate PR

dnephin (approved):
Thank you for the PR! This is a great idea, and it made me think of addition possible addition to `--watch`.

I'm fine with adding a `nolint: gocyclo` when necessary. In this case I was able to move the condition to `handler.runTests`. I think that works out alright, because `runTests` is also the function that sets the value of this field, so keeping both reads and writes in the same place is nice.

Reading over this code again I think a few of the types could probably be renamed (no need for two different things called "handlers"), but I'll leave that for another time.

I also added the `a` key to the docs in the README to make it easier to discover. Assuming CI is happy with the last commit, I will merge.

dprotaso:
> This is a great idea, and it made me think of addition possible addition to --watch.

Another cool watch mode would be to run tests against all dependent packages

For example -  if `C` depends on `B` which depends on `A`. If I change A `gotestsum`  watch will run tests for package A, B and C.

PR #204, opened 2021-07-07
Fix cve-2020-29652

MichaelScheetz-HPE:
golang.org/x/crypto | CVE-2020-29652   | HIGH     | v0.0.0-20200622213623-75b288015ac9


dnephin (approved):
Thank you for the PR!

I believe `gotestsum` never used the code identified in that CVE, but I understand the desire to remove the alert from any scans.

Thankfully it looks like someone did some work to move the one function that required `x/crypto` into `x/term`, and the linter told us where to find it. After running `go mod tidy` the `x/crypto` dependency is gone.

I pushed an additional commit with that change.

MichaelScheetz-HPE:
Thanks for patching that up and merging!

PR #210, opened 2021-08-12
fix(cmd/main.go): add a channel to proc{} for detecting SIGINT

### LINKED ISSUE #209: SIGINT is indistinguishable from a failed test
# Description
I run `gotestsum` in a loop in a bash script. In order to detect SIGINT in the bash script I have to inspect the exit code of `gotestsum` (I think). However,  `<ctrl-c>` during test processing triggers an exit code of 1.

```bash
$ gotestsum . # hit <ctrl-c> immediately
$ echo $?
1
```
The same exit code happens if tests fail.

So, I can't terminate my testing loop based on SIGINT since I want to run all tests - checking if `$? = 1`  will treat failed tests the same way as SIGINT.

I recommend having failed tests keep `$? = 1`. But, OS signals should have `$? > 1`. This makes it easy to predicate on both signals and failed tests.

## Update

I've submitted https://github.com/gotestyourself/gotestsum/pull/210 to address this.

johnrichardrinehart:
Fixes #209 

# Description
This is a solution to fix the usability problem described in https://github.com/gotestyourself/gotestsum/issues/209.

Please provide any suggestions to improve the code quality. Thank you.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"strings"
+	"syscall"
 
 	"github.com/dnephin/pflag"
 	"github.com/fatih/color"
@@ -209,6 +210,12 @@ func run(opts *options) error {
 		return finishRun(opts, exec, err)
 	}
 	exitErr := goTestProc.cmd.Wait()
+	siggedOut := <-goTestProc.signal // check if we received a SIGINT
+
+	if siggedOut != nil {
+		n, _ := (siggedOut).(syscall.Signal)
+		return finishRun(opts, exec, fmt.Errorf("syscall.Signal==%d", int(n)))
+	}
 	if exitErr == nil || opts.rerunFailsMaxAttempts == 0 {
 		return finishRun(opts, exec, exitErr)
 	}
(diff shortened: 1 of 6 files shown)


dnephin (approved):
Thank you for raising this issue and opening a PR to fix it!

Your changes were a great start. I had to make a few changes to get it to work:
* replaced the channel with an `in32` that is accessed atomically. We don't need coordination here, so the atomic int seems like it should work well without having to deal with reading from a channel that could be empty
* changed `startGoTest` to return a pointer to a `proc`, so that the caller would be reading the same field as the signal handler function.
* changed `main()` to accept an exit code from any error, instead of only from an `exec.ExitError`.
* add `128` to the signal number to match bash (https://tldp.org/LDP/abs/html/exitcodes.html)

This seems to work for me, let me know if it is also working for you.

johnrichardrinehart:
@dnephin Awesome! Happy you considered my PR. Please give me 24 hours to review. I'm confused why the atomic load-store was necessary, but I trust that it was. Would you please explain this to me? I figured my channel solution would handle synchronization/races.

Thanks for taking the time to clean this up. I fully expected some revisions, so I'm happy to see your commits :) .

dnephin:
> One question, to start: Did my code fail to compile/execute, for you?

Ya, my comment was misleading I think. I believe it did work, but it felt to me like the atomic was going to be slightly safer and easier to work with.

Thanks again for raising this issue and getting it fixed!

PR #215, opened 2021-08-27
Fix color codes on windows

### LINKED ISSUE #214: ANSI codes printed on Windows
Instead of colored output, ANSI codes are printed as symbols in Windows.
Tested in GitBash, PowerShell, CommandPrompt. All of them produce same incorrect output.

image
image
image

dnephin:
Fixes #214


dnephin:
Thank you for confirming!

PR #224, opened 2021-12-16
fix: remove deprecated linters from golanci-lint

greut:
golangci-lint passes like before.


dnephin (approved):
Thank you!

I made one change, which was to re-add `interfacer`. I understand it is archived, but I still find it to be useful, especially on a small project like this one. Sure sometimes it will give unhelpful suggestions, but I've found that to be rare. I can see it being more of a problem on large projects with many interfaces, but even then I think many of its suggestions are helpful.

PR #226, opened 2021-12-30
Instructions to compile `test2json` without local go

g-gaston:
With the current instructions you actually need a local go installation. Even if the go source is downloaded, a go binary is needed to compile test2json, even if it's a different version than the `test2json` one.

This PR adds instructions to download a temporary go binary, compile test2json from it and delete it.

### DIFF
--- a/docs/running-without-go.md
+++ b/docs/running-without-go.md
@@ -3,15 +3,16 @@
 `gotestsum` may be run without Go as long as the package to be tested has
 already been compiled using `go test -c`, and the `test2json` tool is available.
 
-The `test2json` tool can be compiled from the Go source tree:
+The `test2json` tool can be compiled from the Go source tree so that it can be distributed to the environment that needs it.
 
 ```sh
-export GOVERSION=1.13.5
-curl -Lo go.zip "https://github.com/golang/go/archive/go${GOVERSION}.zip"
-unzip go.zip
-rm -f go.zip
-cd go-go${GOVERSION}/src/cmd/test2json/
-env GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags="-s -w" .
+GOVERSION=1.17.6
+OS=$(uname -s | sed 's/.*/\L&/')
+mkdir -p gopath
+GOPATH=$(realpath gopath)
+HOME=$(realpath ./)
+curl -L --silent https://go.dev/dl/go${GOVERSION}.${OS}-amd64.tar.gz | tar xz -C ./
+env HOME=$HOME GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GOPATH=$GOPATH ./go/bin/go build -o test2json -ldflags="-s -w" cmd/test2json
 mv test2json /usr/local/bin/test2json
 ```
 
@@ -27,4 +28,3 @@ Example: running without a Go installation
 export GOVERSION=1.13
 gotestsum --raw-command -- test2json -t -p pkgname ./binary.test -test.v
 ```
-


dnephin (commented):
Thank you for the PR!

Can you help me better understand the use case that is better served by this change to the docs?

I believe even in the updated version of the docs the `cmd/test2json` path refers to a Go source checkout. Since a Go source checkout is needed in both cases, it seems like it might be easier for most use cases to simply compile all the binaries from the source checkout.

Technically the change is correct. You could use a different version of Go to build `test2json`, but I don't understand in what cases that would be preferable.  I guess it might be faster to download than to build all the binaries?

These instructions are mostly intended to be guidelines, to help someone understand the dependencies needed to run `gotestsum`. I don't expect them to capture every use case. Maybe we could document this as an alternative approach?

g-gaston:
> Thank you for the PR!
> 
> Can you help me better understand the use case that is better served by this change to the docs?
> 
> I believe even in the updated version of the docs the `cmd/test2json` path refers to a Go source checkout. Since a Go source checkout is needed in both cases, it seems like it might be easier for most use cases to simply compile all the binaries from the source checkout.
> 
> Technically the change is correct. You could use a different version of Go to build `test2json`, but I don't understand in what cases that would be preferable. I guess it might be faster to download than to build all the binaries?
> 
> These instructions are mostly intended to be guidelines, to help someone understand the dependencies needed to run `gotestsum`. I don't expect them to capture every use case. Maybe we could document this as an alternative approach?

@dnephin of course!

You are right, the source code is needed in both cases. However, with the current docs, we need a `go` binary already installed and in path in order to compile the `test2json` binary. If we tried to run those commands in a fresh machine without `go`, the command from line 14 would fail. And if the machine already has `go` installed, why not use the second method (which is significantly simpler)? Maybe I'm missing something here and my whole hypothesis is wrong :)

I think I didn't explain correctly my goal. I didn't intend for the docs to show how to compile `test2json` from a different go version, but to show how to compile and install `test2json` in a machine without a previous `go` installation.

If you prefer to not include this in the docs because it's a edge case or because that wasn't the intent of this doc, that's cool with me. I just thought that the idea was to show how to use `gotestsum` without an actual dependency on a `go` local installation. Or if you prefer me to leave the current explanation as is and add a new section, I can also do that 🙂

As for my usecase, I was trying to build an image for running tests in CI but I didn't want to install `go` just to install `test2json`.

By the way, I just updated the PR with a couple env vars I missed that are necessary for `go build`

docs/running-without-go.md
```diff
@@ -3,15 +3,18 @@
 `gotestsum` may be run without Go as long as the package to be tested has
 already been compiled using `go test -c`, and the `test2json` tool is available.
 
-The `test2json` tool can be compiled from the Go source tree:
+The `test2json` tool can be compiled without installing go:
```
dnephin, line 6:
```suggestion
The `test2json` tool can be compiled from the Go source tree so that it can be distributed to the environment that needs it.
```

docs/running-without-go.md
```diff
@@ -3,15 +3,18 @@
...
 
 ```sh
-export GOVERSION=1.13.5
-curl -Lo go.zip "https://github.com/golang/go/archive/go${GOVERSION}.zip"
-unzip go.zip
-rm -f go.zip
-cd go-go${GOVERSION}/src/cmd/test2json/
-env GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags="-s -w" .
+GOVERSION=1.13.5
```
dnephin, line 9:
```suggestion
GOVERSION=1.17.6
```

We may as well update this to something recent while we are in here making changes.

docs/running-without-go.md
```diff
@@ -3,15 +3,18 @@
...
+OS=$(uname -s | sed 's/.*/\L&/')
+mkdir -p gopath
+GOPATH=$(realpath gopath)
+HOME=$(realpath ./)
+curl -L --silent https://go.dev/dl/go${GOVERSION}.${OS}-amd64.tar.gz | tar xz -C ./
+env HOME=$HOME GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GOPATH=$GOPATH ./go/bin/go build -o test2json -ldflags="-s -w" cmd/test2json
+rm -rf go
+rm -rf gopath
```
dnephin, line 17:
```suggestion
```

Let's leave any cleanup out of these instructions so prevent anyone from accidentally deleting something they wanted to keep.

dnephin (commented):
Thanks for the explanation! That's a good point about the old version. I missed that those instructions also required an existing Go binary.

If I remember correctly, the purpose of these docs was to show how to build `test2json` so that it can be copied to another machine for CI. 

I think the edits you made are generally a good improvement, but I think saying "without a Go install" is a bit misleading. I would categorize the `curl` step that downloads Go as installing it. Once Go is downloaded there's no reason to build `test2json` for the local machine, since you could run `go tool test2json`. Building the binary is more for distributing to other places without having to distribute the whole Go toolchain.

I've suggested a few small edits, but otherwise looks great!

dnephin (approved):
LGTM, thank you!

g-gaston:
@dnephin done, PTAL 🙂

PR #234, opened 2022-02-22
Pass stdin to go test

uhthomas:
It may be useful to pipe the `go test` JSON to `gotestsum`. There are a few ways in which this could be achieved:

1. Passing stdin to `testjson.ScanConfig` seems the most sensible, but it may be complicated and may cause unintended side effects. For instance, assuming `gotestsum` were to contextually switch between using stdin and `go test`, existing invocations of `gotestsum` where data is being piped would no longer work. Providing a new flag to opt-in may be helpful? Maybe a sub-command?

2. (This PR) Passing stdin to `--raw-command` allows stdin to be piped to stdout via use of a tool like `cat`.

---

Given the changes included in this PR, `gotestsum` can read `go test` JSON via stdin with:

```sh
cat out.json | gotestsum --raw-command -- cat
```

This may be useful when the JSON is generated via other tools (i.e Bazel). For example:

```sh
bazel test //... |& go tool test2json | gotestsum --raw-command -- cat
# or
bazel test //... |& gotestsum --raw-command -- go tool test2json
```

What do you think?

### DIFF
--- a/cmd/main_test.go
+++ b/cmd/main_test.go
@@ -2,10 +2,12 @@ package cmd
 
 import (
 	"bytes"
+	"encoding/json"
 	"os"
 	"os/exec"
 	"strings"
 	"testing"
+	"time"
 
 	"gotest.tools/v3/assert"
 	"gotest.tools/v3/env"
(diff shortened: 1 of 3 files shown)


cmd/main_test.go
```diff
@@ -399,3 +401,44 @@ func TestRun_RerunFails_PanicPreventsRerun(t *testing.T) {
...
+			Test    string  `json:",omitempty"`
+			Elapsed float64 `json:",omitempty"`
+			Output  string  `json:",omitempty"`
+		}{
+			{Action: "run", Package: "pkg"},
+			{Action: "run", Package: "pkg", Test: "TestOne"},
+			{Action: "fail", Package: "pkg", Test: "TestOne"},
+			{Action: "fail", Package: "pkg"},
+		} {
+			assert.NilError(t, e.Encode(event))
```
dnephin, line 431: I think `t.Fatal` (and anything that results in `t.FailNow`) is only supposed to be called from the main test goroutine. Here `assert.NilError` will call `t.FailNow`.  One way to work around this is to use `t.Error` (`t.Fail`) instead, which can be done with `assert.Check(t, e.Encode(event))`.

dnephin (commented):
Thank you for the PR! I think it would be good to allow some way to read from stdin.

I thought this had come up before, but I was not able to find the discussion, so I'll need to think some more about any potential complications. One thing that comes to mind is that `--watch` changes the terminal mode to accept keystrokes. I'm not sure if that will cause any problems with passing `os.Stdin` to the command being run. You can see that in `internal/filewatcher/term_unix.go`.

dnephin:
Thank you again for the PR! I had a chance to test it out and everything seems to work as expected.  I made a couple small changes to the test to ensure that it would fail when the code change it is testing is removed.  Also added another example to the docs to show how it works. If CI is happy I will merge.

uhthomas:
Thanks! Happy to help. 🙂

PR #235, opened 2022-02-25
Ensure we still rerun tests when there are warnings

### LINKED ISSUE #232: possible bug: build errors passes except when rerun is required
I have a test that has build errors in it with rerun is enabled, which I think causes it to pass, except when there are test errors.  When there are individual test errors, it reports:

```
ERROR rerun aborted because previous run had errors
```

But otherwise, it passes.

It seems inconsistent to me that the build would pass, except when a test needs a rerun.  This makes it hard to know there is a problem until a test fails.

Just for context, the failures are because I'm trying to run with coverage and `--coverpkg` must include only packages that are included in the build, so we get a few errors like this in the error report: 
```
warning: no packages being tested depend on matches for pattern ...
```

Might we want this test to fail immediately rather than only when a rerun is required?

ashanbrown:
Proposal for skipping over warnings.

Elapsed times in golden files are causing this to fail.  It's also a pretty heavy-weight test for such a small change so maybe we should test this some other way.

Fixes #232

### DIFF
--- a/cmd/main_e2e_test.go
+++ b/cmd/main_e2e_test.go
@@ -248,3 +248,29 @@ func TestE2E_MaxFails_EndTestRun(t *testing.T) {
 	)
 	golden.Assert(t, out, "e2e/expected/"+t.Name())
 }
+
+func TestE2E_IgnoresWarnings(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for short run")
+	}
+
+	flags, opts := setupFlags("gotestsum")
+	args := []string{"--rerun-fails=1", "--packages=./testdata/e2e/ignore_warnings/", "--", "-tags=testdata",
+		"-cover", "-coverpkg=github.com/pkg/errors"}
+	assert.NilError(t, flags.Parse(args))
+	opts.args = flags.Args()
+
+	bufStdout := new(bytes.Buffer)
+	opts.stdout = bufStdout
+	bufStderr := new(bytes.Buffer)
+	opts.stderr = bufStderr
+
+	err := run(opts)
+	assert.Error(t, err, "exit status 1")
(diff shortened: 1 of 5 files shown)


dnephin:
Thanks for working on this fix! Sorry it's taken me a bit of time to review.

I think what you have looks like a reasonable way to solve the problem. I'll take a closer look and try to fix the test failure in the next few days.

dnephin (approved):
Looks good! Thank you!

dnephin:
I think the test failures are because the `GOTESTSUM_FORMAT` env var is set in CI, so the output of the test is different. I've updated the test to set a `--format`, and updated the golden file as well.

internal/text/testing.go
```diff
@@ -34,9 +35,8 @@ func OpRemoveSummaryLineElapsedTime(line string) string {
...
 }
 
+var trailingElapsedTimePattern = regexp.MustCompile(`\s*\(\d+(\.\d+)?m?s\)\s*$`)
+
 func OpRemoveTestElapsedTime(line string) string {
-	if i := strings.Index(line, " (0."); i > 0 && i+8 == len(line) {
-		return line[:i]
-	}
-	return line
+	return trailingElapsedTimePattern.ReplaceAllString(line, "")
```
dnephin, line 41: I don't think I see why this change was made. I generally try to avoid regex. Reverting this still has the tests passing. 

I'm going to revert this part of the change for now. We can always re-introduce it later if it fixed something that I'm missing.

ashanbrown: The time was being reported in `ms` rather than seconds so it didn't start with 0.  Sounds like changing `GOTESTSUM_FORMAT` will fix that.

PR #241, opened 2022-04-09
Create any missing directories for {json,junit}file

### LINKED ISSUE #238: ability to force-create path where junit file must be created
Hello,
Thank you for nice tool!

Do you accept PRs? :)
I'd like to make gostest sum force create full path for junit file if given as `--junitfile foo/bar/tests.xml`. 
Whyt, should it be separate option or default behavior?

dnephin:
Closes #238

If the directory can not be created then the file creation operation will fail as it did before.

No reason to force someone to create this directory ahead of time.

### DIFF
--- a/cmd/handler.go
+++ b/cmd/handler.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


glebsts:
Thank you!!!

PR #261, opened 2022-06-27
Add project name to junit.xml output

nikolaigut:
Add attributes `name`, `timestamp`, `tests`, `failures`, `errors`, `skipepd`, `time`, `hostname` to the `testsuites` xml node, according to the schema: https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -88,6 +88,9 @@ func setupFlags(name string) (*pflag.FlagSet, *options) {
 		"format the testsuite name field as: "+junitFieldFormatValues)
 	flags.Var(opts.junitTestCaseClassnameFormat, "junitfile-testcase-classname",
 		"format the testcase classname field as: "+junitFieldFormatValues)
+	flags.StringVar(&opts.junitProjectName, "junitfile-project-name",
+		lookEnvWithDefault("GOTESTSUM_JUNITFILE_PROJECT_NAME", ""),
+		"name of the project used in the junit.xml file")
 
 	flags.IntVar(&opts.rerunFailsMaxAttempts, "rerun-fails", 0,
 		"rerun failed tests until they all pass, or attempts exceeds maximum. Defaults to max 2 reruns when enabled.")
@@ -152,6 +155,7 @@ type options struct {
 	hideSummary                  *hideSummaryValue
 	junitTestSuiteNameFormat     *junitFieldFormatValue
 	junitTestCaseClassnameFormat *junitFieldFormatValue
+	junitProjectName             string
 	rerunFailsMaxAttempts        int
 	rerunFailsMaxInitialFailures int
 	rerunFailsReportFile         string
--- a/internal/junitxml/report.go
+++ b/internal/junitxml/report.go
@@ -17,8 +17,13 @@ import (
 
 // JUnitTestSuites is a collection of JUnit test suites.
 type JUnitTestSuites struct {
-	XMLName xml.Name `xml:"testsuites"`
-	Suites  []JUnitTestSuite
+	XMLName  xml.Name `xml:"testsuites"`
+	Name     string   `xml:"name,attr,omitempty"`
+	Tests    int      `xml:"tests,attr"`
+	Failures int      `xml:"failures,attr"`
+	Errors   int      `xml:"errors,attr"`
+	Time     string   `xml:"time,attr"`
+	Suites   []JUnitTestSuite
 }
 
 // JUnitTestSuite is a single JUnit test suite which may contain many
@@ -64,10 +69,12 @@ type JUnitFailure struct {
 
 // Config used to write a junit XML document.
 type Config struct {
+	ProjectName             string
 	FormatTestSuiteName     FormatFunc
 	FormatTestCaseClassname FormatFunc
 	// This is used for tests to have a consistent timestamp
 	customTimestamp string
+	customElapsed   string
 }
 
 // FormatFunc converts a string from one format into another.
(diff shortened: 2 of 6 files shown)


internal/junitxml/report.go
```diff
@@ -17,8 +17,16 @@ import (
...
-	Suites  []JUnitTestSuite
+	XMLName   xml.Name `xml:"testsuites"`
+	Name      string   `xml:"name,attr,omitempty"`
+	Timestamp string   `xml:"timestamp,attr"`
+	Tests     int      `xml:"tests,attr"`
+	Failures  int      `xml:"failures,attr"`
+	Errors    int      `xml:"errors,attr"`
+	Skipped   int      `xml:"skipped,attr"`
+	Time      float64  `xml:"time,attr"`
+	Hostname  string   `xml:"hostname,attr,omitempty"`
```
dnephin, line 28: I'm not sure this is correct. From my read of the XSD file, these should be properties of `JUnitTestSuite` not `JUnitTestSuites`. Am I missing something? Can you explain more about why they would be on `JUnitTestSuites` ?

`JUnitTestSuite` already has many of these fields, but we could add the ones that are missing.

internal/junitxml/report_test.go
```diff
@@ -19,8 +22,34 @@ func TestWrite(t *testing.T) {
 	out := new(bytes.Buffer)
 	exec := createExecution(t)
 
+	tpl, err := template.ParseFiles("testdata/junitxml-report.golden.tpl")
+	assert.NilError(t, err)
+	f, err := os.Create("testdata/junitxml-report.golden")
+	assert.NilError(t, err)
```
dnephin, line 28: Instead of a templated expected value, I think it is easier to maintain when we patch the expected value. See `Config.customTimestamp` for an example of that. We should be able to specify a value for `hostname` in the test, so that our expected value doesn't change between hosts.

I believe timestamp is already patched, so everything else shouldn't need a template.

That change should allow us to remove the template and go back to a regular golden file.

testjson/execution.go
```diff
@@ -471,6 +472,17 @@ func (e *Execution) Elapsed() time.Duration {
 	return clock.Now().Sub(e.started)
 }
 
+func (e *Execution) ElapsedOnWrite() time.Duration {
```
dnephin, line 475: What's the purpose of this? Does `Elapsed()` not give the correct value for some reason?

dnephin (commented):
Thank you for the PR!  Sorry it took me some time to read over it.

One of the challenges of Junit.xml is that there is no official spec (as far as I've seen). Some people have published their version of a spec, but I don't think the tools and systems that read the file all agree.

So far the `junit.xml` output of `gotestsum` has mostly been limited to what common CI systems will use. It doesn't make any effort to add other fields unless they will be used by something.

Can you elaborate on your use case? What tool or system is receiving the `junit.xml`, and how does it make use of these fields? Understanding that will help ensure we are building the right thing.

I think it would be fine to add some more fields. I've left some suggestions an questions below as well.

internal/junitxml/report.go
nikolaigut: This is a little tricky. In some definitions are these properties available for `JUnitTestSuites` (e.g., https://github.com/junit-team/junit5/blob/main/platform-tests/src/test/resources/jenkins-junit.xsd) and for others not.
I am using this parser to generate a user-friendly view of the tests https://lotterfriends.github.io/online-junit-parser/ and there these properties are used for the title.

cmd/main.go
```diff
@@ -59,6 +59,9 @@ func setupFlags(name string) (*pflag.FlagSet, *options) {
 	flags.StringVarP(&opts.format, "format", "f",
 		lookEnvWithDefault("GOTESTSUM_FORMAT", "short"),
 		"print format of test input")
+	flags.StringVarP(&opts.name, "name", "n",
+		lookEnvWithDefault("GOTESTSUM_NAME", "unit tests"),
+		"the name of the project")
```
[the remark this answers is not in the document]
nikolaigut: I agree with your proposal to change the flag.

nikolaigut:
> Thank you for the PR! Sorry it took me some time to read over it.
> 
> One of the challenges of Junit.xml is that there is no official spec (as far as I've seen). Some people have published their version of a spec, but I don't think the tools and systems that read the file all agree.
> 
> So far the `junit.xml` output of `gotestsum` has mostly been limited to what common CI systems will use. It doesn't make any effort to add other fields unless they will be used by something.
> 
> Can you elaborate on your use case? What tool or system is receiving the `junit.xml`, and how does it make use of these fields? Understanding that will help ensure we are building the right thing.
> 
> I think it would be fine to add some more fields. I've left some suggestions an questions below as well.

Thank you very much for your review.

I need these changes to generate a more user-friendly report with the tool https://lotterfriends.github.io/online-junit-parser/, which uses some of these properties in the title.

internal/junitxml/report_test.go
nikolaigut: Thank you for the example. I changed the test.

internal/junitxml/report.go
dnephin: Even in that jenkins schema, the `testsuites` only has name, time, tests, failures, disabled, and errors. It doesn't have hostname, timestamp, or skipped. Those are all properties of `testsuite`. 

Looking at the source of the tool you linked, 

https://github.com/lotterfriends/online-junit-parser/blob/d279015fac02c402ec12ce033f1a433508914b35/index.html#L122-L132

I don't see hostname in there either. Hostname is on the testsuite (not testsuites).  

So something still doesn't look right here.

I looked for title in that source, and I guess that's the `<h1>` here, is that right?

https://github.com/lotterfriends/online-junit-parser/blob/d279015fac02c402ec12ce033f1a433508914b35/index.html#L174

That seems to only use the name. Are you sure these other properties are required?

testjson/execution.go
dnephin: I'm still curious about this. I think we can find a better way to solve this than introducing two new methods on `Execution`.

dnephin (commented):
Thanks! The test is looking good. Just a few more questions to make sure I understand the purpose of some of these additions.

nikolaigut:
> Thanks! The test is looking good. Just a few more questions to make sure I understand the purpose of some of these additions.

Thanks for this input. You are totally right. I removed the unused properties.

dnephin (approved):
Looks good! Thank you for the PR!

I rebased to resolve the conflicts, squashed the commits, and made a couple small changes 
* renamed the flag to match the other `junitfile` flags, and restored the previous default of no name
* changed the field type of `Elapsed` to `string` to match the other time field

PR #267, opened 2022-08-02
Update Readme

dnephin:
Remove as many files and directories from the top-level to reduce the scroll distance to the README.
Remove the notes about `testjson` library, probably not relevant to most people, and could be confusing.
Move the demo into the "Output Format" section so that the feature list is closer to the top.
Include the flag names in the documentation index, and improve the descriptions.


dnephin:
demo

PR #269, opened 2022-08-06
Test against go1.19, remove go1.15

dnephin:
And update `gofmt` for new go.19 format.

Closes #183

Because go1.19 now includes these fixes in the stdlib: https://tip.golang.org/doc/go1.19#os-exec-path


dnephin:
I'm not entirely sure how to preview the `gofmt` changes, but I'll have to remember to check them out after the next release is tagged.

PR #276, opened 2022-09-03
Add --watch-chdir flag

dnephin:
With the `--watch-chdir` flag, `gotestsum` will change the working directory
to the directory with the modified file before running tests. Changing the
directory is primarily useful when the project contains multiple Go modules.
Without this flag, `go test` will refuse to run tests for any package outside
of the main Go module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,12 @@ packages as command line arguments (ex: `gotestsum --watch -- ./...` or
 `gotestsum --watch -- ./extrapkg`), the
 tests in those packages will also be run when any file changes.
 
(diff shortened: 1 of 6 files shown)


dnephin:
This might not be necessary with the new go workspaces. I guess in most cases creating a worksapce should be fine, but I have at least one use case where it's not.

Go plugins are required to be built with all the same module versions as the binary that loads the plugin. When writing a plugin for `golangci-lint` in a git repo that contains other modules, I want to be able to use different dependencies for each module, but run the tests for any of the modules.

I haven't found a way to do that yet with go worksapces. It forces me to use the shared dependencies, which breaks loading the plugin.

PR #282, opened 2022-11-03
Skip empty tests

### LINKED ISSUE #228: suppress printing package names with no tests
Is there a way to avoid printing packages when only skipped tests or no tests at all are in package?

flowchartsman:
First pass to fix #228

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -184,8 +184,15 @@ func (o options) Validate() error {
 			"because not all test cases will run")
 	}
 	if o.skipEmpty {
-		if len(o.packages) > 0 && o.packages[0] != `./...` {
-			return errors.New("--skip-empty cannot be used with a target")
+		switch len(o.packages) {
+		case 0:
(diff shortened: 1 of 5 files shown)


dnephin:
Thank you for working on this change!

It should be pretty easy to hide empty packages. By definition they don't have any tests, so all we should need to do is avoid printing the "end of package" event in the formatter. There currently isn't a way to pass options to formatters, but I think this is a great time to add options.

Since this is an explicit flag, I don't think we need to handle the package arg in any special way. Checking `./...` is probably the most common case, but you can use the 3 dots on any package (ex: `./somepkg/...` would also be valid and would match `somepkg` and all its subpackages). 

I think we should call the flag `hide` instead of skip, because skip suggests a test exists and is not being run (that's how the `testing` package uses the term). 

I put together those changes in #283 and added you as a co-author. How does that look to you? Does that seem like it solves the problem?

PR #283, opened 2022-11-05
Hide empty packages in compact formats

### LINKED ISSUE #228: suppress printing package names with no tests

dnephin:
Related to #228
Closes #282

Add a new `--format-hide-empty-pkg` flag to hide empty package in compact formats. The `testname` format should be verbose enough that a few extra `EMPTY` lines for packages is not a big deal, and the `standard` formats should emulate the `go test` output as close as possible, so we don't want it to apply to those. 

I tested both formats manually, and added golden tests for `pkgname`.

### DIFF
--- a/cmd/handler.go
+++ b/cmd/handler.go
@@ -57,7 +57,7 @@ func (h *eventHandler) Close() error {
 var _ testjson.EventHandler = &eventHandler{}
 
 func newEventHandler(opts *options) (*eventHandler, error) {
-	formatter := testjson.NewEventFormatter(opts.stdout, opts.format)
+	formatter := testjson.NewEventFormatter(opts.stdout, opts.format, opts.formatOptions)
 	if formatter == nil {
 		return nil, fmt.Errorf("unknown format %s", opts.format)
 	}
--- a/cmd/handler_test.go
+++ b/cmd/handler_test.go
@@ -59,7 +59,7 @@ func (bufferCloser) Close() error { return nil }
 func TestEventHandler_Event_WithMissingActionFail(t *testing.T) {
 	buf := new(bufferCloser)
 	errBuf := new(bytes.Buffer)
-	format := testjson.NewEventFormatter(errBuf, "testname")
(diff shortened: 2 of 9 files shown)


flowchartsman:
This is great, but part of the reason I wrote the original code the way I did was so that I didn't end up with bloated junit.xml files. I have a relatively large project under a single module that has many directories that will never have tests, so it is useful to me not to clutter the junit.xml with a bunch of  empty entries that will always be empty.

Not entirely sure what the right approach is; perhaps instead of the option to hide empty directories, there should be an option to only test those directories which are _testable._

flowchartsman:
Actually, now that I think about it, a  `--junit-hide-empty-pkg` option would probably be the right call, since that would allow people to skip either display _or_ output, selectively.

dnephin:
That sounds like a good approach!

flowchartsman:
Sounds great. Can't wait to see this in the next release. Provided both flags can land, I'm comfortable closing my original PR

dnephin:
I'm not sure when I'll have the time to add `--junit-hide-empty-pkg`. If you're willing to PR the change I'd be happy to review it. I think the approach in this PR should be pretty similar to what is needed for the junit file.

The `junitxml` package already has a `Config` struct, so it should be easy to pass in that setting (without the extra work I had to do here). I'm guessing it would be a check here: https://github.com/gotestyourself/gotestsum/blob/main/internal/junitxml/report.go#L105-L106 to skip adding the package when `pkg.Total == 0`.

There are already some tests for that package using `golden.Assert`. Copying one of those should make it easy to test this change.

flowchartsman:
Sounds good. Feel free to merge this in, and I'll start a separate PR

flowchartsman (approved):
Not sure my approval is necessary, but LGTM :)

dnephin:
Thanks for having a look! Always good to get another set of eyes on a change.

PR #288, opened 2022-12-02
Hide empty packages on junit output

flowchartsman:
Followup to #282 and #283. I am leaving the failed golden test in here, because I think perhaps it might be a bug.  The original golden test output for junit.xml includes the following section, which my code hides:

```xml
<testsuite tests="0" failures="0" time="0.001000" name="gotest.tools/gotestsum/testjson/internal/badmain" timestamp="0001-01-01T00:00:00Z">
	<properties>
		<property name="go.version" value="go7.7.7"></property>
	</properties>
	<testcase classname="" name="TestMain" time="0.000000">
		<failure message="Failed" type="">sometimes main can exit 2&#xA;FAIL&#x9;gotest.tools/gotestsum/testjson/internal/badmain&#x9;0.001s&#xA;</failure>
	</testcase>
</testsuite>
```

This looks wrong to me, since `tests="0"` and `failures="0"`, yet there's clearly a failing testcase being simulated. Not sure if it has to do with the empty classname or if a failure like this is likely to occur or not, but it could conceivably get eaten by a test suite during display.  Not sure how the different libraries handle having both values set to zero and still having a non-skipped test case, but I definitely want to confirm before I tell remove it from the golden file.

Also worth noting: I also tried `len(pkg.Passed) ==0 && len(pkg.Failed)==0 && len(pkg.Skipped)==0` and no change.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -132,3 +132,11 @@ func (s *stringSlice) Set(raw string) error {
 func (s *stringSlice) Type() string {
 	return "list"
 }
+
+func truthyFlag(s string) bool {
+	switch strings.ToLower(s) {
+	case "true", "yes", "1":
+		return true
+	}
+	return false
+}
--- a/cmd/handler.go
+++ b/cmd/handler.go
@@ -96,6 +96,7 @@ func writeJUnitFile(opts *options, execution *testjson.Execution) error {
 		ProjectName:             opts.junitProjectName,
 		FormatTestSuiteName:     opts.junitTestSuiteNameFormat.Value(),
 		FormatTestCaseClassname: opts.junitTestCaseClassnameFormat.Value(),
+		HideEmptyPackages:       opts.junitHideEmptyPackages,
 	})
 }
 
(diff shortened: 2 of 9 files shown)


dnephin (approved):
Thank you for the PR! This looks great!

It was a good call to keep that failing test around. The scenario this is testing is when a package has a `TestMain` function or `init()` function. Those functions can fail or panic, and when they do the count of tests will be 0, but not because the package was empty. The count is 0 because none of the tests in the package ran (due to the earlier failure from `TestMain` or `init`).

I fixed that by checking `Package.TestMainFailed()`. When that is true it's likely the test count is empty because of the earlier failure.

I made a small change to the field name in `junitxml.Config`, to make it consistent with the other field name and the command line flag. I updated the expected value for the help output (to fix a test failure), and adjusted the help text slightly.

I went back and checked the changes from #283, to see if they also need to use this new `IsEmpty` function. Two of  them seemed fine as-is because they only hide output when the package ends in `Skip` or `Pass`. We know there are no `TestMain` failures in those cases (the package will always end with `Fail` if there are `TestMain` failures). Only the `dots-v2` format needed a fix, so I fixed that in another commit, which I also pushed to this PR.

If CI is happy with these changes, I think this is ready to merge.

Issue #289, opened 2022-12-07
Compile+Release build artefacts for FreeBSD aka GOOS=freebsd


ghuntley:
👋 Howdy,

Friendly request and checking in if there's interest on your end to release builds for gotestsum that target FreeBSD to GitHub? `go install gotest.tools/gotestsum@latest && gotestsum` does indeed work on FreeBSD v13x.

# ecosystem notes

Through https://github.com/vmactions/freebsd-vm it is possible to run FreeBSD via GitHub Actions btw. 

<img width="1710" alt="CleanShot 2022-12-08 at 03 43 23@2x" src="https://user-images.githubusercontent.com/127353/206268692-3f6e1eba-e2c8-42b6-bccd-524665a5dcb7.png">

<img width="1012" alt="CleanShot 2022-12-08 at 03 52 12@2x" src="https://user-images.githubusercontent.com/127353/206268666-e0a61d3b-8934-478d-98fa-1bf86a13b622.png">

# related issues
- This all relates to upstream work over at https://github.com/coder/coder/pull/5340

dnephin:
Hello! If someone has a need for it I would be happy to accept a PR that adds freebsd to the list of release binaries. The config for that is here: https://github.com/gotestyourself/gotestsum/blob/main/.project/goreleaser.yml

Which architectures, just amd64?

PR #291, opened 2022-12-08
feat(freebsd): generate arm/amd64 binaries for FreeBSD

ghuntley:
fixes https://github.com/gotestyourself/gotestsum/issues/289

re: exclusion of `arm`

https://www.freebsd.org/platforms/arm/

> 32-bit ARMv6 and ARMv7 is officially a [Tier 2](https://docs.freebsd.org/en/articles/committers-guide/#archs) architecture, as the [FreeBSD](https://www.freebsd.org/) project does not provide official releases or pre-built packages for this platform due to it primarily targeting the embedded arena.
> 
> FreeBSD/arm64 supports 64-bit ARMv8 processors and is a [Tier 1](https://docs.freebsd.org/en/articles/committers-guide/#archs) architecture as of 13.0. 64-bit ARM platforms follow a set of standard conventions, and a single FreeBSD build will work on hardware from multiple vendors. As a result, FreeBSD provides official releases for FreeBSD/arm64 and packages are available.

### DIFF
--- a/.project/goreleaser.yml
+++ b/.project/goreleaser.yml
@@ -10,6 +10,7 @@ builds:
   - binary: gotestsum
     goos:
       - darwin
+      - freebsd
       - windows
       - linux
     goarch:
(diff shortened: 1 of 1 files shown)


dnephin (approved):
Looks good! I see the two new binaries in the output of the "build binaries" step [in CI](https://app.circleci.com/pipelines/github/gotestyourself/gotestsum/616/workflows/981730fe-f31d-48a8-ae97-ef2116fc6551/jobs/4866).

PR #293, opened 2022-12-15
Add option to use high-visibility icons

n-oden:
The default icons used in the pkgname output formats (∅,✓,𐄂) can be hard to spot when scrolling through test output.

This PR adds the ability to pick a higher-visiblity/contrast set of icons (🔳,✅,❌) by adding a `-hivis` suffix to either of the 'pkgname' style formats.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,9 @@ The `--format` flag or `GOTESTSUM_FORMAT` environment variable set the format th
 is used to print the test names, and possibly test output, as the tests run. Most
 outputs use color to highlight pass, fail, or skip.
 
+The `--format-hivis` flag changes the icons used by `pkgname` formats to higher
+visiblity unicode characters.
+
 Commonly used formats (see `--help` for a full list):
 
  * `dots` - print a character for each test.
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -61,6 +61,8 @@ func setupFlags(name string) (*pflag.FlagSet, *options) {
 		"print format of test input")
 	flags.BoolVar(&opts.formatOptions.HideEmptyPackages, "format-hide-empty-pkg",
 		false, "do not print empty packages in compact formats")
+	flags.BoolVar(&opts.formatOptions.UseHiVisibilityIcons, "format-hivis",
+		false, "use high visibility characters in some formats")
 	flags.BoolVar(&opts.rawCommand, "raw-command", false,
 		"don't prepend 'go test -json' to the 'go test' command")
 	flags.BoolVar(&opts.ignoreNonJSONOutputLines, "ignore-non-json-output-lines", false,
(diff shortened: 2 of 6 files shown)


dnephin (commented):
Thank you for the PR! This is great, we should definitely get this merged!

What operation system are you using? I'm wondering how well these characters will work on MacOS, Windows, and Linux, and if anyone would have to install a new font to get it to work.  I can test it on Linux.

If these characters are well supported, I'd like to switch to them by default. Once this merges I'll create an issue and pin it, to have other people test it out before we make it the default. 

There are two things I'd like to think about a bit before we merge:

1. is `hivis` the right term for this? These characters seem to render very differently. Is it because they are part of some unicode character range, or they have some specific property? I don't know much about unicode.
2. should this be enabled with a separate flag `--format-hivis`, or should it be a variant on the format name. I'm thinking since it applies to at least 2 formats (maybe more if we can use this with `dots` and `dots-v2` as well), maybe it should be `--format-hivis` to enable it.

n-oden:
I personally am testing this on MacOS, but the characters are standard unicode/utf-8 codepoints:

🔳 = U+1F533 "[white square button](https://emojipedia.org/white-square-button/)"
✅ = U+2705 "[white heavy check mark](https://emojipedia.org/check-mark-button/)"
❌ = U+274C "[cross mark](https://emojipedia.org/cross-mark/)"

How those particular codepoints render is platform-dependent, but you can see from the emojipedia links above that they are _generally_ similar on most popular platforms.  All three codepoints were adopted as part of Unicode 6.0 in 2010 so at this point you'd have to be running a notably old system for them not to render.

To your specific questions:

1. I'm not overly attached to the name.  "Hivis" is a common abbreviation for "high-visibility" and my goal here was to make test failures stand out more clearly in the pkgname output formats.  As noted above these are at this point pretty bog-standard unicode characters and should be well supported by any terminal program built in the last 10 years without requiring custom fonts or special configuration. 
2. The way I implemented this, you could add the `-hivis` suffix to any of the existing format types, it's just that it's a no-op for any formatter that doesn't use `shortFormatPackageEvent()`. This seemed like the least-inconvenient way to handle it since it meant that the switch statement in `NewEventFormatter()` didn't need to change, but if you'd prefer to do it as a separate flag that's certainly doable.

n-oden:
Happy new year! @dnephin any thoughts on how you'd like to proceed with this?

dnephin:
I tried this out on my desktop and it looks great! The characters render correctly in all the fonts I use on Linux.

I made a few small changes here: https://github.com/odenio/gotestsum/compare/main...dnephin:gotestsum:hivis-icons
* use `--format-hivis` to enable these characters, instead of separate formats. This should make it easier to change to the new characters as the default in the future
* removes the constants and moved the definitions inline, because we only ever reference these characters from one function
* I tried out a different character for "empty package". The "heavy minus" character is in the same unicode code block as the other two characters ("dingbats"), and is slightly less bright, which I'm hoping helps de-emphasize those  packages

Normally I would push those changes to this PR branch, but I do not have permission to do that in this case (maybe you unchecked that box when opening the PR?).

How do you feel about those changes? If they work for you, could you cherry-pick those 2 commits into this branch? Alternatively I can open a PR against your branch, and you can merge that PR to include them here.

n-oden:
@dnephin I merged your commits back up into our main branch, so they're part of this PR now.  👍 on all your changes!

dnephin (approved):
LGTM, Thank you!

PR #298, opened 2023-01-08
Print the test timeout panic as package output

### LINKED ISSUE #292: Panic stack trace from `go test -timeout=X` is ignored when timeout happens right after a test passes
This seems like a bug in `go test -json`, so I've opened the following issue over there: https://github.com/golang/go/issues/57305

However, this is still relevant for `gotestsum` since previous Go versions are affected.

The following visual should give an idea of what's wrong:

```
❯ gotestsum --jsonfile=trace-missing.json -- . -timeout=1s
✖  test (1.012s)

=== Failed
=== FAIL: test  (0.00s)
FAIL	github.com/mafredri/test	1.012s

DONE 1 tests, 1 failure in 1.323s
```

Something went wrong, but we simply don't know what. With parallel tests this is even more confusing:

```
❯ gotestsum --jsonfile=parallel-trace-missing.json -- . -timeout=1s
✖  test (1.008s)

=== Failed
=== FAIL: test TestWorld (unknown)

DONE 2 tests, 1 failure in 1.326s
```

The test fails with `(unknown)`, but no explanation.

When this edge case isn't hit we see the proper output:

```
❯ gotestsum --jsonfile=trace-valid.json -- . -timeout=1s
✖  test (1.01s)

=== Failed
=== FAIL: test TestWorld (unknown)
panic: test timed out after 1s
running tests:
	TestWorld (0s)

goroutine 5 [running]:
[...]

DONE 2 tests, 1 failure in 1.317s
```

The json payloads are available at https://github.com/golang/go/issues/57305, but for context, this is the significant portion:

```
{"Time":"2022-12-14T09:49:02.56982657Z","Action":"output","Package":"github.com/mafredri/test","Test":"TestHello","Output":"--- PASS: TestHello (1.00s)\n"}
{"Time":"2022-12-14T09:49:02.572963923Z","Action":"output","Package":"github.com/mafredri/test","Test":"TestHello","Output":"panic: test timed out after 1s\n"}
{"Time":"2022-12-14T09:49:02.572982687Z","Action":"output","Package":"github.com/mafredri/test","Test":"TestHello","Output":"running tests:\n"}
{"Time":"2022-12-14T09:49:02.572992095Z","Action":"output","Package":"github.com/mafredri/test","Test":"TestHello","Output":"\tTestHello (1s)\n"}
```

What's going on here is that the test outputs `PASS`, relied on by `gotestsum` to ignore future output from that test. However, `go test` considers that test to be running still for a brief moment and attributes the panic stack trace to it.

dnephin:
This is a workaround for the race in `test2json`. See https://github.com/golang/go/issues/57305

Fixes #292

### DIFF
--- a/.project/golangci-lint.yml
+++ b/.project/golangci-lint.yml
@@ -1,6 +1,4 @@
 linters-settings:
-  gocyclo:
-    min-complexity: 12
   goconst:
     min-len: 2
     min-occurrences: 4
@@ -33,7 +31,6 @@ linters:
     - errcheck
     - gocognit
     - goconst
-    - gocyclo
     - gofmt
     - goimports
     - gosimple
--- a/testjson/execution.go
+++ b/testjson/execution.go
@@ -109,6 +109,11 @@ type Package struct {
 	// shuffleSeed is the seed used to shuffle the tests. The value is set when
 	// tests are run with -shuffle
 	shuffleSeed string
+
+	// testTimeoutPanicInTest stores the name of a test that received the panic
+	// output caused by a test timeout. This is necessary to work around a race
+	// condition in test2json. See https://github.com/golang/go/issues/57305.
+	testTimeoutPanicInTest string
 }
 
 // Result returns if the package passed, failed, or was skipped because there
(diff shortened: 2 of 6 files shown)


mafredri (approved):
I can confirm I'm unable to reproduce the issue with these changes, nice work!

Let's hope this is the only case where output is attached to a test that passed. 👍🏻

dnephin:
Thank you confirming the fix!

PR #300, opened 2023-01-29
Fix need to save twice to trigger watch on OSX using Neovim

andersjanmyr:
On OSX (12.6.2) using Neovim (NVIM v0.8.2) the triggering event varies between `fsnotify.Rename` and `Create`.
This fix adds `fsnotify.Rename` to the events triggering a new test run.

*Before*
```
handling event "cors_middleware.go": RENAME
handling event "cors_middleware.go": CREATE

Running tests in ./.
exec: [go test -json ./.]
go test pid: 96495
✓  . (cached)

DONE 71 tests in 0.237s
```

*After*
```
handling event "cors_middleware.go": RENAME

Running tests in ./.
exec: [go test -json ./.]
go test pid: 95707
✓  . (cached)

DONE 71 tests in 0.245s
```

### DIFF
--- a/internal/filewatcher/watch.go
+++ b/internal/filewatcher/watch.go
@@ -236,7 +236,7 @@ type fsEventHandler struct {
 var floodThreshold = 250 * time.Millisecond
 
 func (h *fsEventHandler) handleEvent(event fsnotify.Event) error {
-	if event.Op&(fsnotify.Write|fsnotify.Create) == 0 {
+	if event.Op&(fsnotify.Write|fsnotify.Create|fsnotify.Rename) == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


dnephin:
Thank you for the bug fix! This is definitely something that needs to be fixed.

I think this change should be safe because of the throttling of events, but I'd like to do a bit of testing with vim, vs code, and GoLand to make sure it continues to work correctly with those editors. 

The changes in this PR look great, so as long as this continues to work with those editors I think we should be able to merge this PR.

dnephin (approved):
Thank you for fixing this! These changes work well for me. The throttling of events prevents tests from being run extra times.

Issue #303, opened 2023-02-05
Feat: add optional configurable tests summary


pmalek:
### Use case

I have a set of pretty long integration tests and I want to retain the output from them in pretty much the `standard-verbose` fashion (that's what I use for now on CI).

Since the output is so long and I intend to keep it, it's hard to go through the log and find each individual `PASS: TestXXX (12.34s)` entry.

### Proposed solution

It would be really nice if `gotestsum` could add a toggle (opt-in probably) to print a summary of all tests with their run times.

This way we could use `standard-verbose` format but also get the summary at the end.

### Considered alternatives

Parse `--jsonfile` and use `--post-run-command` with a custom written script that will do the above

dnephin:
Thanks for suggesting this! Using `--jsonfile` and `--post-run-command` to do this is a really interesting approach. It should be possible without a custom script:

```sh
gotestsum \
    --jsonfile tmp.json.log \
    --hide-summary all \
    --post-run-command "gotestsum --format testname --raw-command -- cat tmp.json.log"
```
 
Unfortunately, that does not currently work because the [jsonfile is closed](https://github.com/gotestyourself/gotestsum/blob/875f29492fe9f4123b8489763be916f80ffff872/cmd/handler.go#L50-L53) after `--post-run-command` runs.

I'd like to fix that bug. Once fixed the command above should solve that use case, although it is still quite verbose. It might be interesting to add a new flag to enable that behaviour. 

Another approach would be to run `gotestsum --format testname --jsonfile tests.log`. Print only the test names and failure output, but keep a copy of the full verbose log in a file. From that `--jsonfile` you have a few options:

* print the standard verbose format with `gotestsum --format=standard-verbose --raw-command -- cat tests.log`, possibly as a separate build step
* save either the standard verbose output or the `tests.log` as a CI artifact, so that it can be downloaded later if needed for debugging

pmalek:
> `gotestsum --format=standard-verbose --raw-command -- cat tests.log`

This is pretty neat! Thanks, I wasn't aware of this. The small problem with this is that I'll get all the output from e.g. `TestMain` and coverage, which will bloat the output.

---

As for the workarounds: this is what would work for me:

```
gotestsum \
    --jsonfile tmp.json.log \
    ... more flags ...
jq -c 'select(.Action | match("pass|skip|fail"))' < tmp.json.log > tmp2.json.log
$(GOTESTSUM) --format testname --raw-command -- cat tmp2.json.log
```

[...]

PR #305, opened 2023-02-16
Ensure jsonfile is flushed before running post-run-command

dnephin:
While discussing the proposal in #303 I noticed the jsonfile is closed after running the `post-run-command`. In some cases, I think it's possible the post-run-command would read the file before all the contents of the file are flushed.

The test added in this commit seems to pass without these extra Flush, but from what I can tell there's no guarantee the file is sycned to disk without these new calls. Maybe that test would flake occasionally without these new calls to `Sync` ?

### DIFF
--- a/cmd/handler.go
+++ b/cmd/handler.go
@@ -15,8 +15,13 @@ import (
 type eventHandler struct {
 	formatter testjson.EventFormatter
 	err       io.Writer
-	jsonFile  io.WriteCloser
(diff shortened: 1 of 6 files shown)


dnephin:
Oh, this was working because https://stackoverflow.com/a/64093650/444646. At least on POSIX systems the sync isn't necessary.

Maybe this is necessary for windows? Either way I guess it doesn't hurt.

PR #315, opened 2023-03-29
GitHub actions format

dnephin:
Use https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines to collapse sections of the output.


laurentsenta:
👋 joining in from https://github.com/protocol/.github/issues/563

Apart from the failing expectations in CI, are there any big hurdles to getting this released?

dnephin:
Hello! I think this should be ready to go. The test failure is easy to fix, I've left it in so we have an example of a failing test in the output.

I pushed one bug fix and a couple small changes to the output:
* previously output for passing tests was not available. I change to use a local `output` map which can include the passing test cases
* added a newline after the package line, so now all package are separated by a single newline
* only add a collapsible group if the test case has output. Any test cases without output will have spaces in front to align with the arrows

How does this look to you?

dnephin:
I changed the default back to `pkgname`, but I switched to use `github-actions` format if  `format=testname`. Those formats should be similar enough that no one will be upset the format changed slightly.

I'm happy with this so I'll merge it. If you have any suggestions I'll be waiting a couple weeks to tag a new release. We can always make changes before the release.

PR #325, opened 2023-04-13
Fix coverage detection for go1.20

### LINKED ISSUE #322: Coverage not reported
When I run `gotestsum -- -cover ./...` I don't have any coverage reported anywhere. Is that expected? 

My golang version is 1.20.3.

dnephin:
Fix #322

In go versions 1.19 and earlier the coverage details were always on a separate line. As of go1.20 they are after a tab on the same line as the package name.

### DIFF
--- a/testjson/execution.go
+++ b/testjson/execution.go
@@ -365,8 +365,8 @@ func (p *Package) addEvent(event TestEvent) {
 		p.action = event.Action
 		p.elapsed = elapsedDuration(event.Elapsed)
 	case ActionOutput:
-		if isCoverageOutput(event.Output) {
-			p.coverage = strings.TrimRight(event.Output, "\n")
+		if coverage, ok := isCoverageOutput(event.Output); ok {
+			p.coverage = coverage
 		}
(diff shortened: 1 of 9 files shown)


gtramontina:
Hey there @dnephin! Thanks for the fix! Any idea fo when this can be merged?

dnephin:
This can probably be merged, but I wasn't super happy with the logic, and I was hoping too make it a bit more robust to future changes. I could do that in a future PR. Have you been able to test this change? Does it work for you?

marco-m:
It works fine for me:
```
> gotestsum -- -coverprofile=bin/coverage.out ./...
∅  examples/download
∅  examples/phonex8
✓  . (365ms) (coverage: 31.5% of statements)
```

dnephin:
Sounds good! I'll try to do a bit more testing on this before a release. It's merged now so it should be possible to install from the main branch at least.

marco-m:
Thanks @dnephin !

mitar:
I can confirm master branch works for me on Go 1.20.

PR #326, opened 2023-04-19
Add a multi package line format

zalenskivolt:
Add a couple of formatters that outputs multiple packages per line.

Example output:
```shell
% gotestsum --format multipkg-and-test-fails-time -- -count 1 ./...

 376ms ∅  .
 563ms ∅  cmd/internal/signalhandlerdriver
 621ms ∅  contrib/notify
 626ms ∅  internal/dotwriter (6ms) ∅ log ∅ text
 974ms ✓  cmd/tool/matrix (411ms) ✓ slowest (487ms)
 1.14s ✓  internal/aggregate (527ms) ✓ filewatcher (613ms) ✓ junitxml (662ms)
 1.54s ✓  testjson (712ms)
 11.1s ✓  cmd (10.557s)

% gotestsum --format multipkg2-and-test-fails-time -- -count 1 ./...  # using dotwriter to tabulate all packages

 529ms ∅  .
 11.6s ✓  cmd (10.817s)
 798ms ∅  cmd/internal/signalhandlerdriver
 1.42s ✓  cmd/tool/matrix (626ms) ✓ slowest (538ms)
 798ms ∅  contrib/notify
 1.75s ✓  internal/aggregate (742ms) ∅ dotwriter ✓ filewatcher (832ms) ✓ junitxml (891ms) ∅ log ∅ text
 2.02s ✓  testjson (672ms)

 182 tests, 1 skipped in 11.615s
```


dnephin:
Thank you for your interest in contributing new gotestsum formats!

Looking at the example output, the two changes from the existing `pkgname` format seem to be:

1. grouping of packages on a single line
2. printing the elapsed time at the start of the line

Does that sound right?

What's the motivation for the grouping of packages? It might be difficult to predict how to group packages, as many projects use very different package layouts.

zalenskivolt:
> Thank you for your interest in contributing new gotestsum formats!
>
> Looking at the example output, the two changes from the existing `pkgname` format seem to be:
> 
> 1. grouping of packages on a single line
> 2. printing the elapsed time at the start of the line
> 
> Does that sound right?

Thanks! Yes. I added versions with and without -time. Maybe a separate command line option would be better to control things like this? Same with -and-test-fails for both this formatter and pkgname.

> What's the motivation for the grouping of packages? It might be difficult to predict how to group packages, as many projects use very different package layouts.

It uses a super simple heuristic for this: as long as the next package is another subpackage on the same level, it will combine them (with one additional small special case). I think this might work fairly well for many projects. As soon as there are multiple sub packages on the same level it should pick up many of them. It should always be obvious which package the shortened names refer to, but for clarity it line breaks and outputs the full package name on failure.

My motivation: For the current project I'm working on, this changes the output from some 65+ lines that I have to scroll through, to just one short overview of 20 lines.

It struggles a bit when there are many deeply nested packages. Maybe it could be updated to handle this better.

zalenskivolt:
> It struggles a bit when there are many deeply nested packages. Maybe it could be updated to handle this better.

Updated to allow backing up one step with `↶`, for example:
`✓  pkg/bloom (1.154s) ✓ data/gen (3.949s) ✓ genx (3.949s) ✓ ↶durablequeue (8.301s) ✓ xbytesutil (1.134s)`

dnephin:
I think this is a cool idea! 

Instead of `↶` could we remove the trimming of the shared part of the package path? Every package would be printed with it's relative package path, which removes the need to use `↶` and `.../`. I think that's going to be more obvious to people and shouldn't add too much extra space. It may even reduce the number of lines, because any number of packages can fill the line (instead of using a heuristic). I expect that should simplify the implementation a bit as well.  I guess this new format should still use the dotwriter logic to make sure package names aren't wrapped across lines.

Maybe this new format could be called something like `pkgname-compact`, so that it follows the same naming convention as `pkgname`, `testname` ?  

I like your idea of using a `--format-with-fails` flag to enabling printing failures inline for these package formats. That would let us hide the `pkgname-and-test-fails` format.

Do we need both the time at the start of the line and after the package name? Keeping it only after the package name seems like it works well.

zalenskivolt:
> I think this is a cool idea!

Thanks!

> Instead of ↶ could we remove the trimming of the shared part of the package path? Every package would be printed with it's relative package path, which removes the need to use ↶ and .../. I think that's going to be more obvious to people and shouldn't add too much extra space. It may even reduce the number of lines, because any number of packages can fill the line (instead of using a heuristic). I expect that should simplify the implementation a bit as well. I guess this new format should still use the dotwriter logic to make sure package names aren't wrapped across lines.

At first I read this as just showing the last part of the package name, and this gives a spectacularly short output! Failures still get printed with the full name, so I actually think it might be a useful option. But then I realized you meant printing the full (relative) package paths. This adds quite a bit to the output. For influxdb, many lines fit only two packages for example. But I agree it would be the most obvious format to use by default.

I would love to experiment a bit more around this. I'll add it as a hidden option. I'll also try printing only sibling packages but without any extra line breaks, and just skip all ↶ and .../:s in that case.

> Maybe this new format could be called something like pkgname-compact, so that it follows the same naming convention as pkgname, testname ?

Updated.

> I like your idea of using a --format-with-fails flag to enabling printing failures inline for these package formats. That would let us hide the pkgname-and-test-fails format.

Updated, pkgname-and-test-fails still works but is hidden.

> Do we need both the time at the start of the line and after the package name? Keeping it only after the package name seems like it works well.

I wanted this to see the actual wall-clock time of the messages, from exec.Elapsed(). I updated to use a --wall-time flag, and added support in pkgname and testname as well.

zalenskivolt:
> [short names]  might be a useful option
>
> I would love to experiment a bit more around this. I'll add it as a hidden option. I'll also try printing only sibling packages but without any extra line breaks, and just skip all `↶` and `.../`:s in that case.

Added --format-compact with the options `relative`(default), `short`, `partial`, or `partial-back` (with `↶`). No extra line breaks or `…/`:s, and a little simplified.

I updated the examples at the top. I think controlling the format might be useful for projects of different sizes and package layouts?


> --wall-time flag

I wanted this one to interactively see the time progress, since the test times only give total partial times for lots of things that are running in parallel.

Should this be --format-wall-time to be listed with other format options? Do you think it's useful?


> called something like pkgname-compact

I think both of the two versions might be useful. The dotwriter one (pkgname-compact2) for terminal usage, where it is ok to rewrite lines. And the plain one (pkgname-compact), for CI usage for example.

Maybe this should also be controlled by some option?

zalenskivolt:
> I think both of the two versions might be useful. The dotwriter one (pkgname-compact2) for terminal usage, where it is ok to rewrite lines. And the plain one (pkgname-compact), for CI usage for example.
>
> Maybe this should also be controlled by some option?

I updated to control this by just adding `-plain` to the `--format-compact`.

Also added `--format-compact dots` for a compact dots-v2 summary for each package:
```go
∅ . ✓ cmd (10.602s)···[62] ✓ cmd/tool/matrix (406ms)···[10] ✓ cmd/tool/slowest (486ms)·· ∅ contrib/notify
```

aslafy-z:
Any news on this PR please? Is it ready to review? Thank you!

PR #330, opened 2023-05-11
Detect CI environments for default `no-color` value

silverwind:
Use [`go-supportscolor`](https://github.com/jwalton/go-supportscolor) to detect color support for [CI environments](https://github.com/jwalton/go-supportscolor/blob/198795bbe247f127f76b501948f2f41fc88b8080/supportscolor.go#L213) and more, which among GitHub Actions will enable color support on number of other environments which previously could not display color at all with `gotestsum`.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -214,13 +214,15 @@ var defaultNoColor = func() bool {
 			"GITEA_ACTIONS",
 			"GITHUB_ACTIONS",
 			"GITLAB_CI",
-			"TEAMCITY_VERSION",
 			"TRAVIS",
 		}
 		for _, ciEnvName := range ciEnvNames {
 			if _, exists := os.LookupEnv(ciEnvName); exists {
 				return false
 			}
+		}
+		if _, exists := os.LookupEnv("TEAMCITY_VERSION"); exists {
+			return false
 		}
 		if os.Getenv("CI_NAME") == "codeship" {
 			return false
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/dnephin/pflag"
 	"github.com/fatih/color"
+	"github.com/jwalton/go-supportscolor"
 	"gotest.tools/gotestsum/internal/log"
 	"gotest.tools/gotestsum/testjson"
 )
@@ -201,7 +202,7 @@ func (o options) Validate() error {
 }
 
 var defaultNoColor = func() bool {
-	if os.Getenv("GITHUB_ACTIONS") == "true" {
+	if supportscolor.Stdout().SupportsColor {
 		return false
 	}
 	return color.NoColor
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -221,12 +221,12 @@ var defaultNoColor = func() bool {
 				return false
 			}
 		}
-		if _, exists := os.LookupEnv("TEAMCITY_VERSION"); exists {
-			return false
-		}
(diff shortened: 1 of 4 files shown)


dnephin:
Thank you for the PR! It would be great to enable color by default on any CI systems that support it. I believe color should already be enabled by default in any interactive terminal.

I'd like to avoiding adding another dependency for this. I believe all we need is a few of the lines around here: https://github.com/jwalton/go-supportscolor/blob/198795bbe247f127f76b501948f2f41fc88b8080/supportscolor.go#L213-L215 

Some systems, like CircleCI already do use color, because they properly emulate a terminal.

It's also worth noting that it's always possible to enable color using `--no-color=false`. This change would enable color by default in some environments, but it should have already been possible to enable color in those environments with the command line flag.

silverwind:
> I'd like to avoiding adding another dependency for this

Having it via a dependency ensures that future environments will only need to be supported in one repo instead of hundrets. I see a benefit in doing it via a dependency, even if it does "a bit more" like this one does.

I planned to PR support for a the new Gitea Actions environment into `go-supportscolor`, but currently [it is struggling with 256 colors](https://github.com/go-gitea/gitea/issues/24777), that's why I held off from it.

In case you still don't want `go-supportscolor`, I will alter the PR as per the recommendations, remove the dependency and also include `GITEA_ACTIONS=true` detection for Gitea Actions which can do 16 color fine currently, and I think that is all this module uses.

`--no-color=false`: I searched for an CLI option to force-enable color, but it wouldn't have dawned me that this double negation actually would do the trick. Good to know, but I would suggest a `--color` or `--force-color` option as that is much more obvious.

dnephin:
> I will alter the PR as per the recommendations, remove the dependency and also include GITEA_ACTIONS=true detection for Gitea Actions

That sounds great! Thank you for making this change!

> I would suggest a `--color`  option as that is much more obvious.

I agree it is not very obvious right now. Hiding the `--no-color` option and changing to `--color` would be a great improvement. We can continue to default that flag to `true` when a terminal or supported CI system is detected.

silverwind:
I have reverted the depdendency and have now replicated this block:

https://github.com/jwalton/go-supportscolor/blob/198795bbe247f127f76b501948f2f41fc88b8080/supportscolor.go#L212-L234

cmd/main.go
```diff
@@ -201,8 +201,32 @@ func (o options) Validate() error {
...
+			"TRAVIS",
+		}
+		for _, ciEnvName := range ciEnvNames {
+			if _, exists := os.LookupEnv(ciEnvName); exists {
+				return false
+			}
+		}
+		if _, exists := os.LookupEnv("TEAMCITY_VERSION"); exists {
+			return false
+		}
```
dnephin, line 226: Could this be part of the `ciEnvNames` list above? Is there something different about this one that I'm missing?

dnephin (approved):
Thank you! This looks great.  Sorry it's taking me a while to review. Just one question

cmd/main.go
silverwind: Ah, it should actually be outside the block that checks for `CI`, because that variable is likely not set as per this:

https://github.com/chalk/supports-color/blob/e06f284193939fbc1c75992871cda2b8192557ec/index.js#L114-L128

I'll fix that, thanks for noticing

silverwind: Fixed in https://github.com/gotestyourself/gotestsum/pull/330/commits/3e999571112b51972133b2b8981dd2143161f544.

dnephin (approved):
Thank you!

Issue #341, opened 2023-07-12
Re-running of failed tests does not handle multiple levels of subtests


brycekahle:
If you have multiple levels of nesting for subtests, the re-running of failed tests does not handle re-running properly and will end up running all the levels in-between the root and the subtest.

Example JSON output:
```
{"Package": "pkg", "Action": "run"}
{"Package": "pkg", "Test": "TestParent", "Action": "run"}
{"Package": "pkg", "Test": "TestParent/TestNested", "Action": "run"}
{"Package": "pkg", "Test": "TestParent/TestNested/TestOne", "Action": "run"}
{"Package": "pkg", "Test": "TestParent/TestNested/TestOne", "Action": "pass"}
{"Package": "pkg", "Test": "TestParent/TestNested/TestTwo", "Action": "run"}
{"Package": "pkg", "Test": "TestParent/TestNested/TestTwo", "Action": "fail"}
{"Package": "pkg", "Test": "TestParent/TestNested", "Action": "fail"}
{"Package": "pkg", "Test": "TestParent", "Action": "fail"}
{"Package": "pkg", "Action": "fail"}
```

This should end up running:
```
go test -json -test.run=^TestParent$/^TestNested$/^TestTwo$ pkg
```
but it runs:
```
go test -json -test.run=^TestParent$/^TestNested/TestTwo$ pkg
go test -json -test.run=^TestParent$/^TestNested$ pkg
```
Notice there are *two* errors here:
1. The regexp for the first line should be `^TestParent$/^TestNested$/^TestTwo$` (missing `$` and `^` around second `/`)
2. It should not run `TestParent/TestNested`.

I do not know what is causing error number 1, but I tracked the cause for number 2 down to the `hasSubTestFailed` being incorrect for the in-between levels. I added a test case that fails if added to `testjson/execution_test.go`:

```
func TestScanOutput_WithNestedSubTestFailures(t *testing.T) {
	source := []byte(`{"Package": "pkg", "Action": "run"}
	{"Package": "pkg", "Test": "TestParent", "Action": "run"}
	{"Package": "pkg", "Test": "TestParent/TestNested", "Action": "run"}
	{"Package": "pkg", "Test": "TestParent/TestNested/TestOne", "Action": "run"}
	{"Package": "pkg", "Test": "TestParent/TestNested/TestOne", "Action": "pass"}
	{"Package": "pkg", "Test": "TestParent/TestNested/TestTwo", "Action": "run"}
[...]

PR #343, opened 2023-07-14
rerun-fails: prevent extra runs when a sub-test fails

brycekahle:
Addresses other part of #341

### DIFF
--- a/testjson/execution.go
+++ b/testjson/execution.go
@@ -223,6 +223,14 @@ func (n TestName) Name() string {
 	return string(n)
 }
 
+func (n TestName) Parent() string {
+	idx := strings.LastIndex(string(n), "/")
+	if idx < 0 {
+		return ""
+	}
+	return string(n)[:idx]
+}
+
 func (p *Package) removeOutput(id int) {
 	delete(p.output, id)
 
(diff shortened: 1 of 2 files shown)


testjson/execution.go
```diff
@@ -538,15 +546,41 @@ func (e *Execution) Failed() []TestCase {
...
+	sort.Slice(tcs, func(i, j int) bool {
+		a, b := tcs[i], tcs[j]
+		if a.Package != b.Package {
+			return a.Package < b.Package
+		}
+		return len(a.Test.Name()) > len(b.Test.Name())
+	})
+
+	var result []TestCase //nolint:prealloc
+	var parents = make(map[string]map[string]bool)
```
brycekahle, line 562: Since you just care about the existence in the map, you can use `struct{}` instead of `bool`. Go understands this and it helps reduce memory usage.
```suggestion
	var parents = make(map[string]map[string]struct{})
```

testjson/execution.go
```diff
@@ -538,15 +546,41 @@ func (e *Execution) Failed() []TestCase {
...
 	for _, tc := range tcs {
-		if !tc.hasSubTestFailed {
-			result = append(result, tc)
+		if _, exists := parents[tc.Package]; !exists {
+			parents[tc.Package] = make(map[string]bool)
 		}
+		if parent := tc.Test.Parent(); parent != "" {
+			parents[tc.Package][parent] = true
```
brycekahle, line 568:
```suggestion
			parents[tc.Package][parent] = struct{}{}
```

testjson/execution.go
```diff
@@ -538,15 +546,41 @@ func (e *Execution) Failed() []TestCase {
...
+// tests that have failed subtests. The parent test will always be run when
+// running any of its subtests.
 func FilterFailedUnique(tcs []TestCase) []TestCase {
-	var result []TestCase
+	sort.Slice(tcs, func(i, j int) bool {
+		a, b := tcs[i], tcs[j]
+		if a.Package != b.Package {
+			return a.Package < b.Package
+		}
+		return len(a.Test.Name()) > len(b.Test.Name())
```
dnephin, line 558: With this approach it might be possible to avoid both of these sorts if we iterate over the test cases in reverse order, and then build the results with prepend instead of append. But maybe that relies too much on an assumed order of the data. I need to think about that a bit more.

dnephin (approved):
Thank you for the PR! I realized that substring matching would cause problems when one test name was a prefix of another that wasn't a subtest.

I added a couple examples to the test case. I pushed a commit that uses exact matching of the parent name instead of simply a substring match, which got those test cases passing.

How does this approach look to you? Does this seem correct?

testjson/execution.go
dnephin: I usually would use `struct{}`, but recently a colleague has been trying to convince me that using a `bool` is slightly better because the memory saving of `struct{}` is negligible, and since the use of this `map` is very localized making a mistake of storing a false is unlikely. The readability of bool would then out weight the benefits of a `struct{}`. I thought I'd give it a shot here.

testjson/execution.go
dnephin: I guess this won't work because the fail event of the parent test is later in the output. So we do need this sort.

brycekahle:
> I realized that substring matching would cause problems when one test name was a prefix of another that wasn't a subtest.

I handled the top-level of this and was the reason there was `TestTop` and `TestTop2` in the test cases. I'm realizing it may not have worked for nested though:

testjson/execution.go
brycekahle: > mistake of storing a false is unlikely

Even if you stored a false, it wouldn't matter since you never examine the actual value.

brycekahle: This is your code base, so whatever you prefer is fine.

brycekahle:
> How does this approach look to you? Does this seem correct?

Yes, it seems correct. The approach looks good. Maybe it would be helpful to add a comment explaining the approach?

Issue #352, opened 2023-08-09
dots-v2 is jittery


howardjohn:
Using dots-v2 is very jittery/jumpy.  Its harder to describe than to show:

https://asciinema.org/a/qeanwygqDJaXp6KcpWhfCOQQs

OR

[cast.txt](https://github.com/gotestyourself/gotestsum/files/12307285/cast.txt)

The upload shows it a bit but not the worst part, as it gets worse as more tests are completed - but there is a size limit I was running into. To run the cast.txt do `asciinema play cast.txt`.

Recording was on linux alacritty. I also did gnome terminal and saw similar.

I am planning to do a bit more investigation into this, but figured i would open the issue in the meantime

howardjohn:
There is a semi-standard escape sequence to 'batch' a write. Example:

```
echo -en "\eP=1s\e\x5c"; echo foo; sleep .1; echo bar;  echo -en "\eP=2s\e\x5c"
```

Will write foo and bar in one go. Its not supported in tmux, though, as far as I can tell

howardjohn:
Aha. I found the last piece of the puzzle. We don't need complex terminal features, we just need to stop doing `clear all; write all` and instead do it line-by-line. This way each cell never goes from `text -> blank -> text`. Pretty simple change, I will send a PR in a bit

PR #353, opened 2023-08-09
WIP: dotsv2: improvements to reduce text jitteriness

howardjohn:
For https://github.com/gotestyourself/gotestsum/issues/352

I don't think this is fully ready but its mostly usable as a POC.

Basically the jitteriness comes from two causes:

1. Unlimited height. If we have 1000 packages, for example, and 50 line terminal then we are still trying to write all 1000 packages. This makes it hard for the terminal emulator to keep up
2. Fast writes. We write every time an event comes in, which can be very very often. Like 1000s per second. Again, we cannot keep up.

This PR fixes this by only including the last `<height>` lines (determined from inspecting the terminal) and writing updates on a 100ms timer instead of fixed. 

This is inspired by https://github.com/docker/compose/blob/80856eacafcf96d6f27e74c07e42386fe662f8ef/pkg/progress/tty.go#L50 which has, IMO, a pretty nice progress UI. Note they don't do the height part.

Some things I think would make this better/acceptable to merge:
* The implementation is racy, easy to fix though.
* We could optimize writes to not write if there are no changes.
  * We could optimize to write faster than every 100ms a few times with backoff to 100ms - probably overkill
* We can not bother writing >height lines, since we just throw it away anyways
* We could move to a pull based model, where every 100ms we compute what to write.
* Prioritize what we write. The last N is not ideal -- we could have [1 running test, 100 skipped tests, 1 running test]. The 2 running test lines are most important. As long as NumCPUs < terminal height, this would always ensure we can see all running tests.

Opening for early feedback


dnephin (commented):
Thank you for tracking down these problems and working on the fix! I think the problems you've found definitely explain the bug.

To throttle the output, could we have `dotwriter.Writer.Flush` only write every 100ms? If it is called more often we simply resetting the buffer (no print or call to `clearLines`). That could potentially delay output for longer if a spike of calls to flush is followed by a long delay, but I think most of the time that wouldn't be a problem.  We could handle that with a ticker that calls `Flush`, which I guess would require copying the buffer instead of simply resetting it. I suspect moving the ticker and the locking to `dotwriter.Writer` may simplify things a bit.

Some additional buffering using `bufio` also sounds good.

Limiting line height by dropping the lines that won't be visible seems good to me. That may also be easier to do in `dotwriter.Writer`, I'm not sure.

dnephin:
I assume this was resolved by #368, but if not I'm happy to re-open.

PR #354, opened 2023-08-10
dots-v2: hide cursor during output

howardjohn:
Contributes to https://github.com/gotestyourself/gotestsum/issues/352

Currently, as the new screen is written the cursor jumps around the terminal. This hides it, writes everything else, then returns it.

Prior art:
https://github.com/docker/compose/blob/80856eacafcf96d6f27e74c07e42386fe662f8ef/pkg/progress/tty.go#L160-L162


dnephin (approved):
Thank you!

PR #358, opened 2023-08-12
Fix dots formats with stderr

### LINKED ISSUE #356: dotsv2: Go test stderr clobbers the outpout
Example: https://asciinema.org/a/NpEIGQp4PCvKVrSxfHubhEN2T

You can see for a split second an error is emitted directly to stderr. We then overwrite it, but our "clearLines" attempt it broken since it doesn't account for it. As a result, I end up with a floating `pkg/test` indefinitely.

We will need to capture stderr and display it in the dotswriter most likely? Or something along those lines.

dnephin:
Fixes #356

TODO: test it

### DIFF
--- a/cmd/handler.go
+++ b/cmd/handler.go
@@ -107,6 +107,14 @@ func newEventHandler(opts *options) (*eventHandler, error) {
 		err:       bufio.NewWriter(opts.stderr),
 		maxFails:  opts.maxFails,
 	}
+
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

dnephin:
I haven't been able to reproduce the issue locally. Seems like this should fix it though.

PR #360, opened 2023-08-20
Export elapsed time to the post-run environment

afbjorklund:
For use with `go-test-report` and for completeness


dnephin (approved):
Thank you for the PR! This is a great addition.

I pushed one small change to remove the `SetElapsed` used for testing. There's no way for the external packages to reset the `timeNow` function, which makes `SetElapsed` destructive in tests.

I replaced it with `text.ProcessLines` which is used in a few places to normalize output when testing across package boundaries.

afbjorklund:
Thanks for the fix! Tried hardcoding 1ms first, but then sometimes it took 2ms instead

In retrospect, maybe the variable should have been called `GOTESTSUM_ELAPSED` ?
I was just extending the existing variables, but maybe it is better with the program name.

The idea was to have the other program look for the exported variables, as defaults:
* https://github.com/vakenbolt/go-test-report/issues/61

`gotestsum --jsonfile test_report.json --post-run-command go-test-report`

And then it looked a bit odd with a more generic "TESTS" being hardcoded, for that purpose...
The end result anyway, would be a test_report.html with a proper executionTime and duration.

Issue #116, opened 2020-05-16
Add a cross-platform tool for desktop notifications using --post-run-command


dnephin:
Building on #74 (#76) it would be nice to have a cross-platform way of using desktop notifications. 

I'd like to avoid additional golang dependencies in `gotestsum` and use the `os/exec` approach for running the notification.

A couple options that come to mind:
1. If there are widely used CLI tools (such as `terminal-notifier` in `contrib/notify/notify-macos.go`) we could add a new `cmd/tool/notify` which detects which CLI is available and shells out to the correct CLI
2. We could build a separate binary using something like https://github.com/martinlindhe/notify and make it available to download from the github releases page.

ShiinaOrez:
Good issue, and cross-platform is important, I prefer to option 2, call `notifier` without user permission isn't a good behavior. Build a light and cross-platform notifier is enough.

Issue #74, opened 2019-10-25
Add post-run hook to support sending desktop notifications


dnephin:
Thanks to @glenjamin for the idea.

Instead of publishing notifications directly I'd like to add support for executing a script that is passed all of the necessary details, and can handle any type of notification.

The script should be called with env vars which contain a summary of the run, and a json blob could be passed to stdin to provide full details about any errors.

Such a feature might look something like this:

```
export GOTESTSUM_POST_RUN_SCRIPT=./scripts/notify-test-results.sh
gotestsum
```

Env vars:
```
TEST_TOTAL=100
TEST_FAILED=2
TEST_SKIPPED=5
TEST_ERRORS=0
```

I'm not sure how to encode more details. The junixml format might work, but is probably not the easiest format to integrate. A custom JSON document might work but will be more involved to maintain. 

Maybe env vars will work for the first version.

glenjamin:
When I use these I tend to have my editor window focused, and my terminal running tests hidden - so ideally the notification should get enough data exposed to be able to include some details about the failure:

- files that failed to compile
- the compilation errors
- tests that failed
- files and line numbers of failed tests

dnephin:
Started some work in #76 which I think would enable a basic notification.

File and line numbers is not something that is currently parsed. Test names are easy enough, but the file and line numbers are for the failed checks not for the tests cases themselves.

glenjamin:
I had a little play around with this, had to switch to a shell exec to be able to handle multiple arguments to the hook command well:

https://github.com/gotestyourself/gotestsum/commit/4728f41e79edb14bf85df67426155f7ec0f67f9e

<img width="358" alt="Screenshot 2019-11-03 at 12 45 39" src="https://user-images.githubusercontent.com/151272/68085347-f9da4b00-fe37-11e9-9ab3-13be0b3b44bd.png">
<img width="363" alt="Screenshot 2019-11-03 at 12 45 26" src="https://user-images.githubusercontent.com/151272/68085349-f9da4b00-fe37-11e9-87dc-d73122766833.png">

dnephin:
Thanks for your work on this! I've merged the original PR which adds the new command line flag, and opened #116 for possible next steps.

I'll also open a PR now for adding it to the documentation in the next release.

PR #362, opened 2023-08-21
Add post-run-command notify implementation for linux

afbjorklund:
Using `notify-send` instead of `terminal-notifier`, and icons instead of emojis.

Renamed the "notify-macos" to the proper "notify_darwin", so the build works.

Sample notification:

gotestsum-notify-linux-pass

gotestsum-notify-linux-fail

Issue #116

* https://github.com/gotestyourself/gotestsum/issues/74#issuecomment-549133324


dnephin (commented):
Thank you for the PR! Adding a post run command for linux would be great. I'll try this out myself soon!

afbjorklund:
There is some debugging log spam when invoked, but I kept it the same as the Mac version for now...

```go
        log.Printf("terminal-notifier %#v", args)
```

```go
        log.Printf("notify-send %#v", args)
```

afbjorklund:
In theory this should also work on Windows: http://vaskovsky.net/notify-send/

afbjorklund:
Tested with `GOOS=windows` and Wine + Mono:

wine-pass

wine-fail

https://github.com/vaskovsky/notify-send/releases

dnephin (approved):
Looks great, thanks!

It wasn't working for me at first, because I had to install `notification-daemon`. I realized the stderr was missing, so I was only able to find the problem by running the command myself.

I pushed a commit to add stdout/stderr output from the notify scripts.

PR #361, opened 2023-08-21
Rename TESTS_ELAPSED environment variable

afbjorklund:
Follow-up on:
* https://github.com/gotestyourself/gotestsum/pull/360

I think looks nicer, and works better in the external programs.

Environment: `GOTESTSUM_JSONFILE` `GOTESTSUM_ELAPSED`

PR #363, opened 2023-08-22
Add the time unit to elapsed time duration

afbjorklund:
One final change, add the time unit to the value...

Less guessing, less "is it nano ? or micro ? or seconds"

And it also allows calling `time.ParseDuration` directly

Not to mention it now has the same format as printed.

* #361


afbjorklund:
Now the invocation will be simpler, and get the correct JSON file and Elapsed time automatically:

```
gotestsum --jsonfile test_report.json --post-run-command "go-test-report --output test_report.html"
```

README.md
```diff
@@ -139,7 +139,7 @@ test run has completed. The binary will be run with the following environment
 variables set:
 
 ```
-GOTESTSUM_ELAPSED       # test time elapsed, in seconds
+GOTESTSUM_ELAPSED       # test time elapsed
```
dnephin, line 142:
```suggestion
GOTESTSUM_ELAPSED       # test run time in seconds
```

dnephin (approved):
Looks great, thank you!

PR #359, opened 2023-08-13
Add testdox format

bitfield:
This format shows test names as full sentences, which is useful both for autogenerating documentation ("dox"), and for encouraging programmers to write their test names as sentences describing the behaviour of the unit under test.

The specific method of formatting test names as sentences is that used by https://github.com/bitfield/gotestdox.

PR #366, opened 2023-08-26
Add format-icons flag for nerdfonts instead of unicode

afbjorklund:
https://www.nerdfonts.com/

Icons from the VS Code glyphs:

* https://microsoft.github.io/vscode-codicons/dist/codicon.html

Or from the GitHub octicons:

* https://primer.style/design/foundations/icons (from "[octocat](https://octodex.github.com/)")

Screenshots with [Ubuntu Mono](https://www.programmingfonts.org/#ubuntu):

* unicode

gotestsum-unicode

* nerdfonts

gotestsum-nerdfonts

Other screenshots with [Roboto Mono](https://www.programmingfonts.org/#roboto).


dnephin (approved):
Thank you for the PR! I haven't installed the fonts yet, but the screenshot looks good.

There were some conflicts due to changes from #359. From the discussion on that PR I thought it would be good to add an alternative icon that worked better with screen readers.  Now that we have so many options for icons I thought it would be better to select the icons using a `--format-icons` flag instead of separate boolean flags for each.

I pushed those changes to this PR as part of resolving the conflicts. I should have tried to keep more of those changes for a separate commit after resolve. I'll do that next time. 

How does this look to you?

afbjorklund:
Having shared parameter for icon formats sounds good to me

afbjorklund:
Funny story: when I went to school, the teacher gave me either 𝓡 (right) or ✓ (wrong).

afbjorklund:
The format from "testname", while not strictly icons, also has its merits...

image

```go
		return icons{
			pass: color.GreenString("PASS"),
			skip: color.YellowString("SKIP"),
			fail: color.RedString("FAIL"),
		}.forAction
```

afbjorklund:
And these icons are rather subtle, for higher resolution screens:

gotestsum-emoticons

afbjorklund:
Now up to 10 different icons, if you count the emojis (hivis)

* default
* codicons
* codicons-circle
* octicons
* octicons-circle
* octicons-filled
* emoticons-circle
* emoticons-filled
* text

<details>

</details>

afbjorklund:
Rebased and added proper help and environment variable for the format icons:

gotestsum-formaticons

afbjorklund:
Side note: you can check the result of the "text" icons with a screen reader:

```
gotestsum --format-icons text  --hide-summary all | espeak -k20
```

The `-k20` causes it to indicate the TEST capital letters with using an increased pitch...

You can also add `-x -q`, if you just want to see and not hear the phoneme mnemonics

-----

```
sk'Ip_
sk'Ip_ s,i:;,Emd'i: sl'aS Int'3:n@L sl'aS s'Igna#lh,and@L@dr,aIv3
p'aas_ s,i:;,Emd'i:_:_: k'aSt
p'aas_ s,i:;,Emd'i: sl'aS t'u:l sl'aS m'eItrIks_:_: k'aSt
sk'Ip_ k'0ntrIb sl'aS n'oUtI2f,aI
p'aas_ s,i:;,Emd'i: sl'aS t'u:l sl'aS sl'oUI2st_:_: k'aSt
sk'Ip_ Int'3:n@L sl'aS d'0traIt3
sk'Ip_ Int'3:n@L sl'aS t'Ekst
sk'Ip_ Int'3:n@L sl'aS l'0g
p'aas_ Int'3:n@L sl'aS 'agrI2g,eIt_:_: k'aSt
p'aas_ Int'3:n@L sl'aS dZ'u:nItksm@L_:_: k'aSt
p'aas_ Int'3:n@L sl'aS f'aIlw0tS3_:_: k'aSt
p'aas_ t'EstdZs@n_:_: k'aSt

d'Vn_ w'0nh'VndrI2d@n n'aInti; 'eIt t'Ests
t'u: sk'Ipt In z'i@roU pOInt f'o@ n'aIn t'u: z
```

```
∅  .
∅  cmd/internal/signalhandlerdriver
✓  cmd (cached)
✓  cmd/tool/matrix (cached)
∅  contrib/notify
✓  cmd/tool/slowest (cached)
∅  internal/dotwriter
∅  internal/log
∅  internal/text
✓  internal/junitxml (cached)
✓  internal/aggregate (cached)
✓  internal/filewatcher (cached)
✓  testjson (cached)

DONE 198 tests, 2 skipped in 0.460s
```

afbjorklund:
Final tweak, added alternative.

```
Format icons:
    default                  the original unicode (✓, ∅, ✖)
    alternative              an alternative unicode (✓, ‐, ✗)
    hivis                    higher visibility unicode (✅, ➖, ❌)
    text                     simple text characters (PASS, SKIP, FAIL)
...
```

cmd/main.go
```diff
@@ -145,6 +149,20 @@ Formats:
...
+    hivis                    higher visibility unicode (✅, ➖, ❌)
+    text                     simple text characters (PASS, SKIP, FAIL)
+  * nerdfonts                requires a font from https://www.nerdfonts.com/
+    codicons                   
+    codicons-circle            
+    octicons                   
+    octicons-circle            
+    octicons-filled            
+    emoticons-circle         󰇵 󰇶 󰇸
+    emoticons-filled         󰱱 󰱴 󰱶
```
dnephin, line 164: Adding `text` sounds great! 

I'm concerned now that maybe we have too many options, at least for the initial release. It can be a lot to choose from, and some of them don't look all that different. `default`, `hivis`, `text` are somewhat distinct from each other.

`alternative` isn't much different from `default` and `hivis`, maybe we could leave that one out for now?

For the `nerdfonts`, maybe we could pick just one of each of the `codicons`, `octicons` and `emoticons` variants? That would give us 6 options in total. We can always add more later.

If you had to pick between the default, `-circle`, and `-filled` variants, which do you think we should take? Maybe  `codicons`, `octicons-circle`, and `emoticons-filled` ?

dnephin (commented):
Thanks again! I was away for a bit, so just getting back to this now.

cmd/main.go
afbjorklund: Okay, will pick out some favorites. Possibly rename the finalists.

afbjorklund: gotestsum-icons

I like the codicons because those are the ones I started the PR with, the octicons because they are consistent between the "circle"/outline and the "filled", and the emoticons because they are fun/colorful and whimsical.
The ones that I left out are either too similar to the default ones and not worth the hassle, or they have poor usability and visibility since the color is too vague or the form is not distinct. Added unicode from go-test-report.

Will use some better names, without dashes.

afbjorklund: The idea was that the icons would be set in environment (or configuration?), and wouldn't be in the way.

afbjorklund: Reduced it down to 4, can add text and emoticons later.

afbjorklund: Might want to add some kind of user configuration to gotestsum, so that they can set their own icons.

dnephin (approved):
Awesome!

PR #367, opened 2023-09-07
Add ENV var to toggle `--format-hivis`

jm96441n:
Adds an additional environment variable to toggle on the `--format-hivis` option rather than always need to pass as an argument to the CLI.


jm96441n:
this is probably made redundant by the work done in https://github.com/gotestyourself/gotestsum/pull/366 as that encompasses the env variables as well

afbjorklund:
Right, it moved to `--format-icons` and `GOTESTSUM_FORMAT_ICONS` as a side-effect of adding more icons.

PR #368, opened 2023-09-08
dotsv2: remove jitters entirely

howardjohn:
As best I can tell, this fully fixes https://github.com/gotestyourself/gotestsum/issues/352

While there were a variety of workarounds I tried to fix this, which had some improvement, the real root cause is doing `clear screen; write`. This leaves some time where each cell in the terminal goes from `<something> -> empty -> <something>`, which leads to blinking.

Instead, we can write and clear as we go. This has the same end result, but avoids the intermediate step where we have an empty cell.

This PR is probably not acceptable as-is as it doesn't even compile on windows. Not sure the best steps there, I neither have experience with implementing this in windows nor a machine to test on. We could keep the old implementation for windows, possibly.

### DIFF
--- a/testjson/testdata/format/dots-v2.out
+++ b/testjson/testdata/format/dots-v2.out
@@ -1,1210 +1,1210 @@
-[?25l
-
-   1ms testjson/internal/badmain 
-
+[?25l[0K
+[0K
+   1ms testjson/internal/badmain [0K
+[0K
  0 tests, 1 failure, 1 error
-[?25h[?25l[1A[2K[1A[2K[1A[2K[1A[2K[1A[2K
-
-   1ms testjson/internal/badmain 
-    🖴  testjson/internal/empty 
-
+[0K[?25h[?25l[5A[0K
+[0K
+   1ms testjson/internal/badmain [0K
+    🖴  testjson/internal/empty [0K
+[0K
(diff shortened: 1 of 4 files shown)


dnephin:
Nice! This works great for me. I also don't have a windows machine. I think leaving the old implementation for windows would be fine.

howardjohn:
> Nice! This works great for me. I also don't have a windows machine. I think leaving the old implementation for windows would be fine.

SGTM, I left it as-is for now and commented that it could be improved in the future if someone wants

testjson/testdata/format/dots-v2.out
```diff
@@ -1,1210 +1,1210 @@
...
+   1ms testjson/internal/badmain [0K
+[0K
  0 tests, 1 failure, 1 error
-[?25h[?25l[1A[2K[1A[2K[1A[2K[1A[2K[1A[2K
-
-   1ms testjson/internal/badmain 
-    🖴  testjson/internal/empty 
-
+[0K[?25h[?25l[5A[0K
+[0K
```
dnephin, line 7: I think it's pretty cool that you can see the functional change to the logic by looking at the golden file. The control sequences moved to a different place in the output.

dnephin (approved):
Thank you making these improvements!

PR #373, opened 2023-10-08
Add support for gotestsum on illumos/amd64

szaydel:
Please consider adding support for illumos-based operating systems such as OmniOS, OpenIndiana, SmartOS, etc. I tested to make sure things build correctly and validated that the tool works correctly on a couple of versions of OmniOS. Addition of `internal/filewatcher/term_illumos.go` was necessary to correctly build, and I apologize, but these constants are not available from `golang.org/x/sys/unix`, thus I supplied literal values instead. These values have not changed in at least 20 years, likely much longer.

### DIFF
--- a/.project/goreleaser.yml
+++ b/.project/goreleaser.yml
@@ -13,6 +13,7 @@ builds:
       - freebsd
       - windows
       - linux
+      - illumos
     goarch:
       - amd64
       - arm64
@@ -35,6 +36,14 @@ builds:
       - goos: windows
         goarch: s390x
       - goos: windows
+        goarch: ppc64le
+      - goos: illumos
+        goarch: arm
+      - goos: illumos
+        goarch: arm64
+      - goos: illumos
+        goarch: s390x
+      - goos: illumos
         goarch: ppc64le
 
 checksum:
--- a/internal/filewatcher/term_illumos.go
+++ b/internal/filewatcher/term_illumos.go
@@ -1,0 +1,4 @@
+package filewatcher
+
+const tcGet = 0x540d
+const tcSet = 0x540e


.project/goreleaser.yml
```diff
@@ -36,6 +37,14 @@ builds:
...
       - goos: windows
         goarch: ppc64le
+      - goos: illumos
+        goarch: arm
+      - goos: illumos
+        goarch: arm64
+      - goos: illumos
+        goarch: s390x
+      - goos: illumos
+        goarch: ppc64le
```
dnephin, line 47: I don't know much about this OS. Are all of these architectures common, or is there 1 that is more common than others?

dnephin (commented):
Thank you for the PR! I think adding the `term_illumos.go` changes would be great so that it builds on illumOS.

I'm a little hesitant to add 4 more build targets to the `goreleaser.yml` for platforms that are much less common.

dnephin (approved):
Oh, this is the ignore list! It only builds on one. This looks good to me

.project/goreleaser.yml
szaydel: Thank you for looking at the PR right away.

These four groups are the ignored combinations, because currently only `illumos/amd64` is actually used. That's the only combo that we'd want to target.

szaydel:
> Oh, this is the ignore list! It only builds on one. This looks good to me

Yeah, that's what I was just trying to say. Thanks for taking a second look.

szaydel:
Thank you again for merging this change in, I appreciate it!

internal/filewatcher/term_illumos.go
```diff
@@ -0,0 +1,4 @@
+package filewatcher
+
+const tcGet = 0x540d
+const tcSet = 0x540e
```
smoynes, line 4: You are almost certainly right that these constants likely won't change ever, but I wonder if you can't use `unix.TCGETS` and `unix.TCSETS`, respectively? If nothing else, having a name makes it a bit less mysterious from where these values originate.

These constants do seem to be present in `golang.org/x/sys/unix` when `GOOS=illumos`, at least with recent versions.

```console
$ GOOS=illumos go doc golang.org/x/sys/unix.TCGETS
...
	TCGETS                        = 0x540d
...
```

szaydel:
Yes, nearly zero chance of changing, but you are right using these named
constants makes sense. I somehow did not see them. If they won't require a
go.mod update I will happily switch to them. Magic numbers are never great.
I will double check and do a PR when I have a chance. Thanks again for the
suggestion!

szaydel:
@smoynes, thanks again. I put up a PR with the change you recommended after making sure that things look correct. https://github.com/gotestyourself/gotestsum/pull/375

PR #374, opened 2023-10-11
Improve warning message when stdin is not a terminal

smoynes:
As discussed in a [comment](https://github.com/gotestyourself/gotestsum/issues/370#issuecomment-1724656425), update the warning message to be slightly more useful to the user by mentioning the impact of the error rather than just the cause.

Because `go test` captures Stderr & Stdout and redirects Stdin, it provides a handy repro. Without this change:

```console
$ go test
Running tests in ./.
Watching 1 directories. Use Ctrl-c to to stop a run or exit.
WARN failed to put terminal (fd 0) into raw mode: operation not supported by device
```

With this change:

```console
$ go test
Watching 1 directories. Use Ctrl-c to to stop a run or exit.
WARN no terminal input -- keyboard shortcuts disabled: operation not supported by device
```

I am happy to change the text if the maintainers can suggest better wording.

Refs: #370

🤵🏻

### DIFF
--- a/internal/filewatcher/term_unix.go
+++ b/internal/filewatcher/term_unix.go
@@ -26,7 +26,8 @@ func newTerminal() *terminal {
 }
 
 // Start the terminal is non-blocking read mode. The terminal can be reset to
-// normal mode by calling Reset.
+// normal mode by calling Reset. If os.Stdin is not a terminal or cannot use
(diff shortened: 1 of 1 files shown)


dnephin (approved):
This is great, thank you!

PR #375, opened 2023-10-12
Used named constants instead of magic numbers in illumos-specific code

szaydel:
This is based on feedback from @smoynes. I validated that these constants are correctly selected with `GOOS=illumos` this makes the code more consistent with other already existing OS-specific code in the `filewatcher` package.

### DIFF
--- a/internal/filewatcher/term_illumos.go
+++ b/internal/filewatcher/term_illumos.go
@@ -1,4 +1,6 @@
 package filewatcher
 
-const tcGet = 0x540d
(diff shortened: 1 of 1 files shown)


smoynes (approved):
FWIW LGTM +1

szaydel:
Thank you for the review and feedback. Cheers!

dnephin (approved):
Thank you!

szaydel:
Thank you @dnephin! Cheers.

PR #377, opened 2023-10-22
Update gotestsum retries to properly filter out parents when there is a a missing gap in the parent tree

ddworken:
If someone creates a test name with `/` in the name, then it is possible to end up with a hierarchy with a gap. For example, if someone creates a subtest of `TestParent` with the name `child/grandchild` and it fails, then the two failing tests will be `TestParent` and `TestParent/child/grandchild`. This breaks `FilterFailedUnique` and will cause it not to filter out `TestParent` which means that all child tests will get rerun. This PR fixes this by updating the `parents` to include all the parents of the given test case. 

If you're curious for why someone would be doing this, I don't think there is any strong requirement for people to do this, but [personally I do it in hishtory](https://github.com/ddworken/hishtory/blob/9433bd6e944c4b59d487e740d271c1e4a63b5bb5/client/client_test.go#L62-L63) and I suspect I'm not the only person doing this. So it seems like a useful change. WDYT?

### DIFF
--- a/testjson/execution.go
+++ b/testjson/execution.go
@@ -563,8 +563,15 @@ func FilterFailedUnique(tcs []TestCase) []TestCase {
 		if _, exists := parents[tc.Package]; !exists {
 			parents[tc.Package] = make(map[string]bool)
 		}
-		if parent := tc.Test.Parent(); parent != "" {
-			parents[tc.Package][parent] = true
+		currentTest := tc.Test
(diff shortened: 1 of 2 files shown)


dnephin (approved):
Thank you for the bug fix! This is great!

I made a minor change to simplify the for loop a bit, but functionally this should be the same thing.

PR #379, opened 2023-10-28
Add Woodpecker CI to no-color default autodetection

xoxys:
[Woodpecker CI](https://github.com/woodpecker-ci/woodpecker) is a fork of Drone and also supports colors in the CI output.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -220,7 +220,7 @@ func defaultNoColor() bool {
 	// true for many CI environments which support color output. So instead, we
 	// try to detect these CI environments via their environment variables.
 	// This code is based on https://github.com/jwalton/go-supportscolor
-	if _, exists := os.LookupEnv("CI"); exists {
+	if value, exists := os.LookupEnv("CI"); exists {
 		var ciEnvNames = []string{
 			"APPVEYOR",
 			"BUILDKITE",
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -230,6 +230,7 @@ func defaultNoColor() bool {
 			"GITHUB_ACTIONS",
 			"GITLAB_CI",
 			"TRAVIS",
+			"woodpecker",
```
dnephin, line 233: This will look for an environment variable `woodpecker=...`, looking at https://woodpecker-ci.org/docs/usage/environment#built-in-environment-variables I think what you want instead is to remove this line and do something like the following after the "codeship"` line below:

```go
if os.Getenv("CI_SYSTEM_NAME") == "woodpecker" {
    return false
}
```

dnephin (commented):
Thank you for the PR! I would be happy to accept a change, but I think this may not work correctly.

cmd/main.go
xoxys: Thanks for the fast review. Sorry, I have not read the code carefully enough. I've changed the code to check the value of the `CI` env var now. If you prefer to use a dedicated lookup, I can also change it to:

```
if os.Getenv("CI") == "woodpecker" {
    return false
}
```

dnephin (approved):
Thank you!

PR #387, opened 2024-02-15
Implemented combine coverprofile in the case of rerun failed

yihuaf:
As discussed in #274, when `gotestsum` rerun failed test cases, the `coverprofile` gets overridden by the each calls to the underlying `go test`, causing coverprofile to per in correct.

I tested the code with the repros provided in #274: https://github.com/neiser/gotestsum-rerun-dependent-subtests/tree/master

The code is just a proof of concept. If the general approach is acceptable, then I will go ahead and clean up the PR with necessary unit tests and comments.


yihuaf:
Looks like `test-windows` is broken unrelated to my change. I tried tip of the tree and the test still fails.

```
=== Failed
=== FAIL: cmd TestE2E_RerunFails/first_run_has_errors,_abort_rerun (0.39s)
    main_e2e_test.go:69: assertion failed: 
        --- expected
        +++ actual
        @@ -1,6 +1,11 @@
        +FAIL testjson/internal/broken
        +
        +=== Failed
        +=== FAIL: testjson/internal/broken 
        +FAIL   gotest.tools/gotestsum/testjson/internal/broken [build failed]
         
         === Errors
         ../testjson/internal/broken/broken.go:5:21: undefined: somepackage
         
        -DONE 0 tests, 1 error
        +DONE 0 tests, 1 failure, 1 error
```

internal/coverprofile/covermerge.go
```diff
@@ -0,0 +1,95 @@
+package coverprofile
+
+// gocovmerge takes the results from multiple `go test -coverprofile` runs and
+// merges them into one profile
+
+// Taken from: https://github.com/wadey/gocovmerge @ b5bfa59ec0adc420475f97f89b58045c721d761c
```
dnephin, line 6: I think we may need to copy the license file
(https://github.com/wadey/gocovmerge/blob/master/LICENSE) into `internal/coverprofile` to comply with the license.

cmd/rerunfails.go
```diff
@@ -51,23 +62,37 @@ func rerunFailsFilter(o *options) testCaseFilter {
 	return testjson.FilterFailedUnique
 }
 
+// Need to add the cov file context and etc...
```
dnephin, line 65: Is this a TODO that still needs to be resolved?

cmd/rerunfails.go
```diff
@@ -83,6 +108,25 @@ func rerunFailed(ctx context.Context, opts *options, scanConfig testjson.ScanCon
 			if exitErr != nil {
 				nextRec.lastErr = exitErr
 			}
+
+			// Need to wait for the go test command to finish before combining
+			// the coverprofile files. But need to before checking for errors.
+			// Even if there is errors, we still need to combine the
+			// coverprofile files.
+			if isCoverprofile {
+				if err := coverprofile.Combine(mainProfile, newCoverprofile); err != nil {
```
dnephin, line 117: It might be better to wait until all the attempts are complete and then combine all the files. That way we aren't re-reading the main file each time.

 Would that work?

dnephin (commented):
Thank you for the PR! I appreciate the contribution.

I had a quick look and overall it sounds reasonable to me. I agree the test failure seems unrelated.

I will find some more time to review in more detail. Some test coverage would also be appreciated!

cmd/rerunfails.go
yihuaf: No. I will remove.

cmd/rerunfails.go
yihuaf: Yes, this would work and better.

yihuaf: After some more thoughts, I implemented the logic as follow:

- each rerun will write the coverage profile into it's own file
- after each rerun, read the rerun coverage profile into memory and delete the file
- after all rerun is done before success, we combine all the rerun coverage profile with the main profile

This achieve 2 goals:
- Only reading the main coverage profile once.
- The coverage is valid upto the latest failed rerun and the rerun cover profile file is cleaned up

PTAL and let me know what you think.

internal/coverprofile/covermerge.go
yihuaf: I added the LICENSE to the file header since it is only the file we copied over. Please let me know if this is good enough.

gaby:
@dnephin I was trying to do a PR to update your dependencies but tests are failing locally with the same error posted by @yihuaf

yihuaf:
@dnephin Let me know if there is anything else you would like me to fix :)

yihuaf:
@dnephin Ping again :)

corani:
@dnephin I'm looking forward to this change as well, to replace my custom script that simply reruns the entire test-suite in case of flaky tests.

PR #403, opened 2024-05-10
fix: close FailsReportFile

testwill:


dnephin (approved):
Thank you for the fix!

PR #406, opened 2024-05-13
go junitxml: compute per package start times rather than using per execution

zdylag:
This can be pretty meaningful if people want to build a timeline of when what test packages were executing. One example use case would be building a trace view ala https://docs.datadoghq.com/tests/setup/junit_xml/?tab=linux.

### DIFF
--- a/testjson/execution.go
+++ b/testjson/execution.go
@@ -131,6 +131,21 @@ func (p *Package) TestCases() []TestCase {
 	return tc
 }
 
+// EarliestTime returns the earliest start time found within a child test case.
+func (p *Package) EarliestTime() time.Time {
+	found := false
+	earliest := time.Time{}
+
+	for _, tc := range p.TestCases() {
+		if !found || tc.Time.Before(earliest) {
+			earliest = tc.Time
+			found = true
+		}
+	}
+
+	return earliest
+}
+
 // LastFailedByName returns the most recent test with name in the list of Failed
 // tests. If no TestCase is found with that name, an empty TestCase is returned.
 //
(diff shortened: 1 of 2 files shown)


testjson/execution.go
```diff
@@ -131,6 +131,23 @@ func (p *Package) TestCases() []TestCase {
 	return tc
 }
 
+// EarliestTime returns the earliest start time found within a child test case,
+// and a bool for if any such time was found.
+func (p *Package) EarliestTime() (time.Time, bool) {
+	found := false
+	t := time.Time{}
```
dnephin, line 138: A more descriptive name for `t` would be great. Maybe `earliest` ?

testjson/execution.go
```diff
@@ -131,6 +131,23 @@ func (p *Package) TestCases() []TestCase {
...
+
+	for _, tc := range p.TestCases() {
+		if !found || tc.Time.Before(t) {
+			t = tc.Time
+		}
+
+		found = true
+	}
+
+	return t, found
```
dnephin, line 148: I think we don't need the bool return value here. Checking if the returned value `IsZero` should be fine.

dnephin (commented):
Thank you for the PR! This fix sounds good

testjson/execution.go
zdylag: I actually originally did that, but thought it was more likely to get flagged in the PR 😅 Sounds good I will change both of these 👍

zdylag:
Hey @dnephin ! Sorry for the super slow turnaround, but I finally addressed your feedback 😅 No rush clearly, but just letting you know it's been updated.

dnephin:
Thank you for the PR! While reviewing this again I realized there are other events that wouldn't be captured in `TestCases` that could include an earlier timestamp.   If I run `go test -json`, I see an `Action=Start` event for packages.


I kept your commit, and I opened #482 to use any event time for both the start and end time. This way time reported by the CLI will match the time reported by junitxml.

If that works for you I can tag a release with the change.

PR #415, opened 2024-06-15
WIP: support reading stdout/stderr from streams

pohly:
It can be useful to let the user run the test command(s) itself, for example when complex shell scripts are involved which need to invoke `go test` multiple times.

With -stdin, the `go test` stdout is expected on stdin of gotestsum, so it can be used in a pipe. To detect abnormal termination of the test commands, bash with "set -o pipefail" should be used. Beware that such failures are not detected by gotestsum.

To also capture stderr with gotestsum, stderr must get redirected like this:

   mkfifo /tmp/pipe
   go test ... 2>/tmp/pipe | gotestsum -stdin -stderr 3 3</tmp/pipe


cmd/main.go
```diff
@@ -70,6 +71,10 @@ func setupFlags(name string) (*pflag.FlagSet, *options) {
 		"use different icons, see help for options")
 	flags.BoolVar(&opts.rawCommand, "raw-command", false,
 		"don't prepend 'go test -json' to the 'go test' command")
+	flags.BoolVar(&opts.readStdin, "stdin", false,
+		"don't run any command, instead read go test stdout from stdin")
+	flags.IntVar(&opts.readStderrFD, "stderr", 0,
```
pohly, line 76: Should this be a named file instead?

Perhaps some elaborate invocation could work without a named pipe. In practice, I think with bash one has to rely on one.

pohly:
I added documentation and unit tests. I also checked that this really works for Kubernetes: https://github.com/kubernetes/kubernetes/pull/125534.

cmd/main_e2e_test.go
```diff
@@ -275,3 +275,104 @@ func TestE2E_IgnoresWarnings(t *testing.T) {
...
+	opts.args = flags.Args()
+
+	bufStdout := new(bytes.Buffer)
+	opts.stdout = bufStdout
+	bufStderr := new(bytes.Buffer)
+	opts.stderr = bufStderr
+
+	in := `{"Time":"2024-06-16T14:46:00.343974039+02:00","Action":"start","Package":"example.com/test"}
+{"Time":"2024-06-16T14:46:00.378597503+02:00","Action":"run","Package":"example.com/test","Test":"TestSomething"}
+{"Time":"2024-06-16T14:46:00.378798569+02:00","Action":"pass","Package":"example.com/test","Test":"TestSomething","Elapsed":0}
```
pohly, line 364: The linter is unhappy about the line length.

Is that a hard criteria for code? Any suggestion how to avoid it? Splitting up the line in the middle IMHO would be worse than a long line.

PR #448, opened 2024-11-02
fix(validate): fixes check for test.failfast flag as this is a gotest…

aldiesel:
Our pipeline would sometimes pass without checking all tests as we had reruns and fail fast enabled.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -208,8 +208,8 @@ func (o options) Validate() error {
 			"when go test args are used with --rerun-fails " +
 				"the list of packages to test must be specified by the --packages flag")
 	}
-	if o.rerunFailsMaxAttempts > 0 && boolArgIndex("failfast", o.args) > -1 {
-		return fmt.Errorf("-failfast can not be used with --rerun-fails " +
+	if o.rerunFailsMaxAttempts > 0 && boolArgIndex("test.failfast", o.args) > -1 {
+		return fmt.Errorf("-test.failfast can not be used with --rerun-fails " +
 			"because not all test cases will run")
 	}
 	return nil
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -208,8 +208,10 @@ func (o options) Validate() error {
 			"when go test args are used with --rerun-fails " +
 				"the list of packages to test must be specified by the --packages flag")
 	}
-	if o.rerunFailsMaxAttempts > 0 && boolArgIndex("test.failfast", o.args) > -1 {
-		return fmt.Errorf("-test.failfast can not be used with --rerun-fails " +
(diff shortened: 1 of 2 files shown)


cmd/main_test.go
```diff
@@ -85,8 +85,8 @@ func TestOptions_Validate_FromFlags(t *testing.T) {
 		},
 		{
 			name:     "rerun-fails with failfast",
-			args:     []string{"--rerun-fails", "--packages=./...", "--", "-failfast"},
-			expected: "-failfast can not be used with --rerun-fails",
+			args:     []string{"--rerun-fails", "--packages=./...", "--", "-test.failfast"},
+			expected: "-test.failfast can not be used with --rerun-fails",
```
dnephin, line 89:
```suggestion
			args:     []string{"--rerun-fails", "--packages=./...", "--", "-failfast"},
			expected: "-failfast can not be used with --rerun-fails",
		},
		{
			name: "rerun-fails with test.failfast",
			args:     []string{"--rerun-fails", "--packages=./...", "--", "-test.failfast"},
			expected: "-test.failfast can not be used with --rerun-fails",
```

cmd/main.go
```diff
@@ -208,8 +208,8 @@ func (o options) Validate() error {
 			"when go test args are used with --rerun-fails " +
 				"the list of packages to test must be specified by the --packages flag")
 	}
-	if o.rerunFailsMaxAttempts > 0 && boolArgIndex("failfast", o.args) > -1 {
-		return fmt.Errorf("-failfast can not be used with --rerun-fails " +
+	if o.rerunFailsMaxAttempts > 0 && boolArgIndex("test.failfast", o.args) > -1 {
+		return fmt.Errorf("-test.failfast can not be used with --rerun-fails " +
```
dnephin, line 212:
```suggestion
	if o.rerunFailsMaxAttempts > 0 && boolArgIndex("failfast", o.args) > -1 {
		return fmt.Errorf("-failfast can not be used with --rerun-fails " +
		"because not all test cases will run")
	}
	if o.rerunFailsMaxAttempts > 0 && boolArgIndex("test.failfast", o.args) > -1 {
		return fmt.Errorf("-test.failfast can not be used with --rerun-fails " +
```

dnephin (commented):
Thank you for the PR! 

I believe both `failfast` and `test.failfast` are valid flag names for `go test`. Instead of replacing the `failfast` check, let's add a second check for the other flag name.

cmd/main_test.go
aldiesel: ahh good call, tested locally as well I took the return from go test for what flags to use literally. Hope the change works

dnephin (approved):
Thank you!

PR #469, opened 2025-02-26
Include build output from Go 1.24

evanphx:
Go 1.24 now has the build errors and such in the json rather than on stderr. This makes sure that those messages show up.


dnephin:
Thank you for the PR! I'm going to cherry-pick parts of this and #474 to see if I can get all the tests passing.

dnephin:
Merged as part of #476. If this works for you I can tag a release.

evanphx:
Sounds good! Thanks for all the work!

Issue #264, opened 2022-07-18
Add attribute `file` to JUnit XML report to accommodate CircleCI algorithm for splitting tests


yterradas:
Hi,

CircleCI allows to split tests by timings using a JUnit XML report.  Unfortunately it requires `testcase` to have an `file` attribute in order to properly work as detailed in this [CircleCI forum question](https://discuss.circleci.com/t/split-by-timings-on-elixir-project-fails-to-find-stored-timings-possibly-missing-file-names-in-junit-xml/34813).  I have confirmed it by manipulating the JUnit XML report output of `gotestsum`.  I understand that `file` is not part of the JUnit XSD but it would really help if it could be added to the output if the user wanted it.  What are your thoughts?

dnephin:
Thanks for opening this issue! The JUnit XML report should definitely work with CircleCI. This is a [blog post from CircleCI](https://circleci.com/blog/level-up-go-test-with-gotestsum/) that covers usage.

With Go, tests can only be split reliably by package because `go test` allows you to easily run tests by package, but running individual tests within the page is much harder (requires very special handling of `-run`).

So CircleCI should use the filename on `testsuite`, instead of `testcase` to split the files. There's an example of that [here](https://github.com/hashicorp/consul/blob/f6edc37d0c18f4c315dfa5aed40a58f254cfce68/.circleci/config.yml#L132). There are many projects using the JUnit xml from `gotestsum` to support test splitting on CircleCI, so I'm pretty confident it works.

You may be missing the `--timings-type=classname` flag to make it work?

yterradas:
I read the post (dated 2018) while investigating the issue.  I have no issue generating the report.  However as stated in the forum post i linked (date 2020) the issue is that CircleCI ignores `classname` if `file` is not present.  By the time I opened this issue I had already tested using `classname`.  I confirmed it doesnt work, as the forum post states.  Only after adding `file` attribute did CircleCI test splitting algorithm worked. 

- logs from worker 0
```
++ circleci tests split --split-by=timings --timings-type=classname --show-counts
Read 53 lines of classname(s)
No timing found for "test/folder/XXX_test.go"
No timing found for "test/folder/XXX_test.go"
...
Bucket 1: assigning 6 classname(s), total weight 32024621988
````
- logs from worker 6
```
++ circleci tests split --split-by=timings --timings-type=classname --show-counts
Read 53 lines of classname(s)
No timing found for "test/folder/XXX_test.go"
[...]

PR #470, opened 2025-02-27
Add the file/line to the output of Junit XML

asfaltboy:
Implement the proposed solution in https://github.com/gotestyourself/gotestsum/issues/264#issuecomment-1189308758 to parse the text output of a failure (which contains a useful trace) add then file/line attributes to the junit xml output.

This is useful when parsing the XML output e.g in CI, in order to link to the file (e.g Github annotations, [as done by mikepenz/action-junit-report](https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#sample-%EF%B8%8F)).

I tested it on a local project, and can confirm the relative path to the file and the line was added as attributes when I ran gotestsum in the root of my module. Here is some anonymised output example excerpt:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="30" failures="4" errors="0" time="17.950271">
	<testsuite tests="30" failures="4" time="14.719000" name="github.com/mycool/package" timestamp="2025-02-27T18:26:48Z">
		<properties>
			<property name="go.version" value="go1.23.2 darwin/arm64"></property>
		</properties>
		<testcase classname="github.com/mycool/package" name="TestSomeService_ProcessFoo_Completed" time="1.170000" file="mycool/package/foo_test.go" line="353">
			<failure message="Failed" type="">=== RUN   TestSomeService_ProcessFoo_Completed ... some long output ... /workspace/mycool/package/foo_test.go:56&#xA;        &#x9;Error:      &#x9;Not equal: &#xA;        &#x9;            &#x9;expected: &#34;ended&#34;&#xA;        &#x9;            &#x9;actual  : &#34;active&#34;&#xA;        &#x9;            &#x9;&#xA;        &#x9;            &#x9;Diff:&#xA;        &#x9;            &#x9;--- Expected&#xA;        &#x9;            &#x9;+++ Actual&#xA;        &#x9;            &#x9;@@ -1,2 +1,2 @@&#xA;        &#x9;            &#x9;-(model.FooStatus) (len=5) &#34;ended&#34;&#xA;        &#x9;            &#x9;+(model.FooStatus) (len=6) &#34;active&#34;&#xA;        &#x9;            &#x9; &#xA;        &#x9;Test:       &#x9;TestSomeService_ProcessFoo_Completed&#xA;--- FAIL: TestSomeService_ProcessFoo_Completed (1.17s)&#xA;</failure>
		</testcase>
		<testcase classname="github.com/mycool/package" name="TestSomeService_CancelFoo_Ok" time="1.070000"></testcase>
		<testcase classname="github.com/mycool/package" name="TestSomeService_CancelFoo_CannotCancel/completed" time="0.020000"></testcase>
	</testsuite>
</testsuites>
```

Merging this would be a proper fix to #264

**TODO**
- [x] correct the relative path (use ~`goModuleFilePath`~ current working dir and clean it up, hard-code in tests), as it's currently a bit too low down the tree

### DIFF
--- a/cmd/tool/parser/parser.go
+++ b/cmd/tool/parser/parser.go
@@ -6,8 +6,7 @@ package parser
 import (
 	"bufio"
 	"fmt"
-	"os"
-	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -17,31 +16,21 @@ import (
 // ParseFailure parses the output of the `go test` for a test failure  and
 // returns the file and line number of the failed test case.
 func ParseFailure(output string) (file string, line int, err error) {
+	re, err := regexp.Compile(`^\s*([_\w]+\.go):(\d+):`)
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to compile regexp: %v", err)
+	}
+
 	scanner := bufio.NewScanner(strings.NewReader(output))
 	for scanner.Scan() {
 		outputLine := scanner.Text()
 		// Usually the failure would contain a line like this:
-		// Error Trace:	/Users/user/proje/path/to/package/some_test.go:42
-		// where the full path to the file is in the same line as "Error Trace:"
-		if strings.Contains(outputLine, "Error Trace") {
-			absolutePath := strings.TrimSpace(strings.TrimPrefix(strings.TrimSpace(outputLine), "Error Trace:"))
-			currentPath, err := os.Getwd()
-			if err != nil {
-				return "", 0, fmt.Errorf("failed getting current path: %v", err)
-			}
+		// some_test.go:42 (surrounded by white-space)
+		// the full path to the file is not available
+		matches := re.FindStringSubmatch(outputLine)
 
-			relPathRaw, err := filepath.Rel(currentPath, absolutePath)
-			if err != nil {
-				log.Debugf("failed to get relative path from trace: %v", err)
-				return "", 0, err
-			}
-			// in case we're deeply nested, remove any repeating dots
-			relPath := filepath.Clean(strings.TrimLeft(relPathRaw, "./"))
-			parts := strings.Split(relPath, ":")
-			if len(parts) != 2 {
-				log.Debugf("failed to split the trace path: %s", relPath)
-				return "", 0, nil
-			}
+		if len(matches) == 3 {
+			parts := matches[1:]
 			file = parts[0]
 			line, err = strconv.Atoi(parts[1])
 			if err != nil {
--- a/cmd/tool/parser/parser_test.go
+++ b/cmd/tool/parser/parser_test.go
@@ -8,13 +8,13 @@ import (
 
 func TestParseFailure_Ok(t *testing.T) {
 	// given
-	failure := `Error Trace:	/project/path/to/package/some_test.go:42`
+	failure := `			 some_1s_test.go:42:  \n`
 
 	// when
 	file, line, err := ParseFailure(failure)
 
 	// then
 	assert.NilError(t, err)
-	assert.Equal(t, file, "project/path/to/package/some_test.go")
+	assert.Equal(t, file, "some_1s_test.go")
 	assert.Equal(t, line, 42)
 }


asfaltboy:
@dnephin kindly review when you get the chance 🙏  happy to address any concerns

internal/junitxml/report.go
```diff
@@ -229,3 +236,12 @@ func write(out io.Writer, suites JUnitTestSuites) error {
 	_, err = out.Write(doc)
 	return err
 }
+
+func appendFailFileLine(jtc *JUnitTestCase, failureOutput string) {
+	file, line, err := parser.ParseFailure(failureOutput)
+	if err != nil {
+		log.Warnf("Failed to parse failure output: %v", err)
```
dnephin, line 243: How about something like this?

```suggestion
		log.Warnf("Failed to parse file:line from test output: %v", err)
```

cmd/tool/parser/parser.go
```diff
@@ -0,0 +1,55 @@
...
+			file = parts[0]
+			line, err = strconv.Atoi(parts[1])
+			if err != nil {
+				log.Debugf("failed to convert line number to int: %v", err)
+				return "", 0, nil
+			}
+			break
+		}
+	}
+	return file, line, err
```
dnephin, line 54: I think this needs to return `scanner.Err()`:

```suggestion
	return file, line, scanner.Err()
```

cmd/tool/parser/parser.go
```diff
@@ -0,0 +1,55 @@
...
+// ParseFailure parses the output of the `go test` for a test failure  and
+// returns the file and line number of the failed test case.
+func ParseFailure(output string) (file string, line int, err error) {
+	scanner := bufio.NewScanner(strings.NewReader(output))
+	for scanner.Scan() {
+		outputLine := scanner.Text()
+		// Usually the failure would contain a line like this:
+		// Error Trace:	/Users/user/proje/path/to/package/some_test.go:42
+		// where the full path to the file is in the same line as "Error Trace:"
+		if strings.Contains(outputLine, "Error Trace") {
```
dnephin, line 26: I don't think this I've ever see an "Error Trace" line like this before. Is this from a specific test helper library?

Most test failures I see look like this:

```
fails_test.go:50: failed sub
```

I don't think we can rely on this prefix.

cmd/tool/parser/parser.go
```diff
@@ -0,0 +1,55 @@
...
+			absolutePath := strings.TrimSpace(strings.TrimPrefix(strings.TrimSpace(outputLine), "Error Trace:"))
+			currentPath, err := os.Getwd()
+			if err != nil {
+				return "", 0, fmt.Errorf("failed getting current path: %v", err)
+			}
+
+			relPathRaw, err := filepath.Rel(currentPath, absolutePath)
+			if err != nil {
+				log.Debugf("failed to get relative path from trace: %v", err)
+				return "", 0, err
```
dnephin, line 36: Instead of `log.Debugf` we probably want to return `fmt.Errorf(..., err)`

dnephin (commented):
Thank you for the PR! If you pull in the latest changes from `main` that should fix the test failures.

In general I think the approach of parsing the file:line from the output is ok, but I don't think it's safe to expect the "Error Trace" prefix.

We may want to do this parsing on the event before we join all the lines. I think that will make it easier to find the relevant `file:line`.

There's also this proposal to add an `OutputType` field: https://github.com/golang/go/issues/62728#issuecomment-1814145946

I think that field would make it much easier to find the appropriate `file:line` in the output. We could parse it directly from only the events that have an `OutputType: error`.

samloop:
Hi, bumping this PR - also a user of gotestsum, and interested in a file attribute for tests. @asfaltboy are you planning on coming back to this?

cmd/tool/parser/parser.go
asfaltboy: no longer relevant

internal/junitxml/report.go
asfaltboy: addressed in https://github.com/asfaltboy/gotestsum/commit/937e59c6c6e49792492ef7a78aa84188b95e5fa6

PR #482, opened 2025-03-28
Use event times for elapsed

dnephin:
Closes #406

This PR uses event times, instead of wall clock time, to calculate the start and elapsed time of packages. In some cases this may report less time because I think compile time was previously included where as now I suspect it is not included. I expect in most cases this is not a problem because someone can easily capture the process runtime outside of gotestsum.

This also fixes the elapsed time reported when `gotestsum` is run from a previously captured test2json event log. Previously it would report runtime of processing that log, which was often 0s and misleading.

Also updates the junitxml to use this event time instead of process start time.


zdylag:
Thanks for this! Really appreciate the changes! 🎉

PR #490, opened 2025-04-14
feat: Add flag to disable `GITHUB_ACTIONS` override

clabby:
## Overview

Adds a new `--format-disable-gha` flag to optionally disable the `github-actions` formatter override when running with `--format testname` or `--format short-verbose` in GitHub actions.

An existing workaround for this is to manually set `GITHUB_ACTIONS=false`, but this is a bit hacky. This PR makes disabling this functionality explicit, and adds some documentation to make this behavior less confusing for the next person.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,11 @@ Commonly used formats (see `--help` for a full list):
 Have an idea for a new format?
 Please [share it on github](https://github.com/gotestyourself/gotestsum/issues/new)!
 
+> [!NOTE]
+> When running with the `testname` or `standard-verbose` format, if the `GITHUB_ACTIONS` environment variable is
+> set, the format will automatically be changed to `github-actions` by default. To prevent this, pass the
+> `--format-disable-gha` flag.
+
 #### Demo
 
 A demonstration of three `--format` options.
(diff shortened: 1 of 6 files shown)


dnephin:
Hello, thanks for the PR! 

Is there a problem with the github actions format? It should be a pretty direct upgrade of the `testname` format. I'd like to better understand the motivation for disabling it.

clabby:
@dnephin Our reason for wanting to disable it is primarily to reduce the size of the test logs. For our integration test suite, loading the GHA log output can take a few minutes as several hundreds of thousands of lines get streamed onto the page. Disabling it speeds up viewing the job output quite a bit for this case.

For the majority of usecases, though, it's a great improvement. Kept the default behavior as most people likely do want the "fold"-style output.

dnephin:
Thanks for the explanation! I'm still a bit hesitant to take this approach given that you can accomplish the same thing with:

```
GITHUB_ACTIONS=  gotestsum ...
```

Maybe another option would be hide the output of successful tests? That's similar to the default behaviour of the `pkgname-and-test-fails`. Maybe that would allow us to remove that option and replace it with this new flag.

Something like `--format-hide-passing-output` ?

BoomchainLabs (approved):

BoomchainLabs (approved):

PR #492, opened 2025-04-16
fix(junit): Correctly identify panicking test in JUnit report

vtsachev:
When a test panics, the JUnit XML report previously marked all tests in the package as failed. This change introduces tracking for the specific test ID that first emits panic output. The JUnit reporter now uses this ID to mark the panicking test with <failure message="Panic"> instead of the generic "Failed", while allowing other tests in the package to retain their correct status. This addresses the issue where a single panic would incorrectly attribute failure to the entire package in the report.

### DIFF
--- a/internal/junitxml/report.go
+++ b/internal/junitxml/report.go
@@ -189,8 +189,14 @@ func packageTestCases(pkg *testjson.Package, formatClassname FormatFunc) []JUnit
 
 	for _, tc := range pkg.Failed {
 		jtc := newJUnitTestCase(tc, formatClassname)
+		msg := "Failed"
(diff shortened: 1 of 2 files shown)


dnephin:
Thanks for the PR! 

Since all the tests in a package run in the same process, I don't think it's safe to assume anything about which test caused the panic. From what I remember the `go test` runner itself is not all that reliable about reporting where the panic came from. Plenty of  times it will parse the panic as not being part of any test, but only part of the package. 

Is the current behaviour causing problems? Do you see panics that often? Maybe there's a way the source of the panics can be fixed so that tests fail instead of panic?

PR #497, opened 2025-05-12
feat: add the `--rerun-fails-abort-on-data-race` flag

hugodutka:
This PR implements the `--rerun-fails-abort-on-data-race` flag. When a race condition is detected, the flag stops re-running failed tests and fails the entire test run.

My motivation for adding this flag is to let `gotestsum` work with race tests that should fail when a data race is found but retry when a test fails for other reasons. In my specific use case, I run a CI test suite where some tests fail due to transient issues, such as network errors; those can be retried and do not require an engineer’s attention. A race condition, however, signals a bug in concurrent code and should notify an engineer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,9 @@ To avoid re-running tests when there are real failures, the re-run will be
 skipped when there are too many test failures. By default this value is 10, and
 can be changed with `--rerun-fails-max-failures=n`.
 
+You may use the `--rerun-fails-abort-on-data-race` flag to abort the re-run if
+a data race is detected.
+
 Note that using `--rerun-fails` may require the use of other flags, depending on
 how you specify args to `go test`:
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -114,6 +114,8 @@ func setupFlags(name string) (*pflag.FlagSet, *options) {
 
 	flags.IntVar(&opts.rerunFailsMaxAttempts, "rerun-fails", 0,
 		"rerun failed tests until they all pass, or attempts exceeds maximum. Defaults to max 2 reruns when enabled")
+	flags.BoolVar(&opts.rerunFailsAbortOnDataRace, "rerun-fails-abort-on-data-race", false,
+		"do not rerun tests if a data race is detected")
 	flags.Lookup("rerun-fails").NoOptDefVal = "2"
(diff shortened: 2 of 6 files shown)


dnephin (commented):
Thank you for the PR!

I think you can accomplish this behaviour today, according to the go docs: https://go.dev/doc/articles/race_detector#Options

I think something like this should work:

```
GORACE="halt_on_error=1" gotestsum ... -- -race ./...
```

You can also set a specific exit code if you want. `gotestsum --rerun-fails` will already stop re-runs if it sees any exit code other than 1.

Would  that work for your use case? I think we could document that to make it easier to discover.

hugodutka:
Thank you for taking a look @dnephin.

Using `halt_on_error` didn't cause `gotestsum` to abort retries on a test with a data race when I tried it:

```
$ GORACE="halt_on_error=1" gotestsum --packages="./scripts/race/..." --rerun-fails=2 -- -count=1 -race
✖  scripts/race (14ms)

DONE 1 tests, 1 failure in 0.259s

✖  scripts/race (13ms)

DONE 2 runs, 2 tests, 2 failures in 0.550s

✖  scripts/race (16ms)

=== Failed
=== FAIL: scripts/race TestRace (unknown)
==================
WARNING: DATA RACE
Write at 0x00c000212238 by goroutine 8:
[...rest truncated...]
```

> You can also set a specific exit code if you want

I believe `GORACE="exitcode=x"` changes the exit code of the test binary instead of go test itself, and as [you noticed before](https://github.com/golang/go/issues/45508), go test overrides it. When I tried using it, gotestsum retried the tests as usual. Let me know if you had a different method of overriding the exit code in mind.

In the same vein, I believe `GORACE="halt_on_error=1"` also applies only to the test binary.

The only workaround I found which lets me detect if `gotestsum` ran into a data race without scanning its output is to use `GORACE="log_path=race.log"`, and then scan the repository for files matching the pattern. This however requires a separate script to run after `gotestsum` finishes, and `gotestsum` will retry the tests anyway.

dnephin:
Ah, that is unfortunate. But maybe this use case is a more compelling reason for them to fix that bug. I'll try opening a new issue when I get a chance.

dnephin (approved):
Thank you!

PR #498, opened 2025-05-12
feat: add support for running failed tests first based on previous ru…

shay-dahan:
…n JSON file

- Introduced a new flag `--failed-first` to specify the path to a JSON file containing the results of the previous test run.
- Implemented logic to read the JSON file and prioritize running failed tests before executing all tests.
- Added unit tests to validate the new functionality and ensure correct behavior when handling various scenarios, including no failed tests and invalid file paths.
- Updated help text to include the new flag description.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Launch Package",
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
+            "args": ["--rerun-fails"],
+            "program": "${workspaceFolder}/main.go"
+        }
+    ]
+}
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
dnephin, line 1: I think this file was committed by accident? Let's leave launch config up to the individual developer. If you'd like to add something consider adding it to `contrib/vscode/launch.json.sample`.

cmd/main.go
```diff
@@ -270,6 +279,71 @@ func run(opts *options) error {
...
+		for _, line := range strings.Split(string(data), "\n") {
+			if strings.TrimSpace(line) == "" {
+				continue
+			}
+			var evt struct{ Package, Test, Action string }
+			_ = json.Unmarshal([]byte(line), &evt)
+			if evt.Action == "fail" && evt.Test != "" {
+				failedTests = append(failedTests, evt)
+			}
+		}
```
dnephin, line 299: There's code in the `testjson` package to do this: https://pkg.go.dev/gotest.tools/gotestsum@v1.12.2/testjson#ScanTestOutput

I'd suggest using that to read the JSON.

dnephin (commented):
Thank you for the PR!

Can you tell me more about the problem this solves for you? Is this to run tests in CI, or running them locally as you make changes?

PR #499, opened 2025-05-12
feat: add --failed-first flag to run failed tests first

shay-dahan:
This update introduces a new command-line flag, --failed-first, which allows users to prioritize running failed tests first. The flag is integrated into the setupFlags function and is accompanied by a new test case to validate its functionality.


dnephin:
Thanks for the PR. I don't understand what you're trying to do here. Run which failed tests before which other tests?

The current diff doesn't do anything, it sets a variable from a command line flag but never uses it.

dnephin:
I guess this PR was opened by accident, and #498 has the changes you want?

PR #515, opened 2025-07-28
fix: graceful exit using ctrl+C if not test are running

### LINKED ISSUE #508: Graceful Exit for `--watch` Mode
### Feature Request: Dedicated Key for Graceful Exit in `--watch` Mode

Currently, when `gotestsum` is running in `--watch` mode, the only way to exit the process is via a `Ctrl+C` interruption. While `Ctrl+C` is a standard way to interrupt a process, it results in a **non-zero exit code**, which doesn't offer a "graceful" way to exit the program.

I propose adding a key like `q` or `e` that allows users to perform a graceful exit from the `--watch` mode. This would provide a cleaner and more intentional way to stop the process.

If that makes sense for the project, I'm interested in contributing to this feature. I'm a beginner in Go and would appreciate any advice or guidance on implementing this feature.

guerinoni:
This will return exit code 0 if gotestsum is waiting for any changes and
it is interrupted, otherwise if some tests are running using ctrl+c it
interrputs the tests and exit with status 1.

Closes #508


cmd/main.go
```diff
@@ -32,14 +32,17 @@ func Run(name string, args []string) error {
...
 	switch {
 	case opts.version:
 		fmt.Fprintf(os.Stdout, "gotestsum version %s\n", version)
 		return nil
 	case opts.watch:
-		return runWatcher(opts)
+		return runWatcher(ctx, cancel, opts)
 	}
-	return run(opts)
+	return run(ctx, cancel, opts)
```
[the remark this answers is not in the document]
guerinoni: Thanks for pointing out!

I did more refactoring a went deep in context propagation and I ended up with a cleaner solution :)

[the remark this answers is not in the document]
guerinoni: Updated the description of PR also

cmd/main.go
```diff
@@ -32,14 +32,32 @@ func Run(name string, args []string) error {
 	opts.args = flags.Args()
 	setupLogging(opts)
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, os.Interrupt)
```
ccoVeille, line 39: I know you are moving code, but you could consider using signal.NotifyContext that would simplify this code

ccoVeille (approved):

cmd/watch.go
```diff
@@ -11,17 +11,16 @@ import (
...
+	w := &watchRuns{opts: *opts, ctx: ctx, cancel: cancelFunc}
 	return filewatcher.Watch(ctx, opts.packages, opts.watchClear, w.run)
 }
 
 type watchRuns struct {
 	opts     options
 	prevExec *testjson.Execution
+	ctx      context.Context
+	cancel   context.CancelFunc
 }
```
ccoVeille, line 24: Wait, there is something strange.

Context should never be embedded in a struct.

I feel like there is something strange

cmd/watch.go
```diff
@@ -53,7 +52,7 @@ func (w *watchRuns) run(event filewatcher.Event) error {
 	opts.packages = append(opts.packages, event.Args...)
 
 	var err error
-	if w.prevExec, err = runSingle(&opts, dir); !IsExitCoder(err) {
+	if w.prevExec, err = runSingle(w.ctx, w.cancel, &opts, dir); !IsExitCoder(err) {
```
ccoVeille, line 55: If you need to pass a context here it means the method you are in should receive it.

cmd/watch.go
```diff
@@ -62,10 +61,7 @@ func (w *watchRuns) run(event filewatcher.Event) error {
...
 // possible to share runSingle with run, but the defer close on the handler
 // would require at least 3 return values, so for now it is a copy.
-func runSingle(opts *options, dir string) (*testjson.Execution, error) {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
+func runSingle(ctx context.Context, cancel context.CancelFunc, opts *options, dir string) (*testjson.Execution, error) {
 	if err := opts.Validate(); err != nil {
 		return nil, err
 	}
```
ccoVeille, line 67: I feel like the code is wrong here

So I wouldn't have updated runSingle signature the way you do.

The idea is not not cancel the context you receive, but to cancel the context you are passing to code you call from this method.

Here is what it should be

```go
func runSingle(ctx context.Context, opts *options, dir string) (*testjson.Execution, error) { 
         ctx, cancel := context.WithCancel(ctx) 
         defer cancel() 
  
          if err := opts.Validate(); err != nil { 
                  return nil, err 
          }
// …
}
```

This implies to refactor the code you just updated.

ccoVeille (changes requested):

cmd/watch.go
guerinoni: Mmmm maybe make sense, this way it will be canceled only this sub-context

cmd/watch.go
ccoVeille: My point was to pass ctx as first argument of this method

guerinoni: Ok fixed :)

cmd/main.go
guerinoni: Nice, I took this way

cmd/main.go
```diff
@@ -464,9 +464,8 @@ func startGoTest(ctx context.Context, dir string, args []string) (*proc, error)
 	}
 	log.Debugf("go test pid: %d", cmd.Process.Pid)
 
-	ctx, cancel := context.WithCancel(ctx)
 	newSignalHandler(ctx, cmd.Process.Pid, &p)
-	p.cmd = &cancelWaiter{cancel: cancel, wrapped: p.cmd}
+	p.cmd = &cancelWaiter{wrapped: p.cmd}
```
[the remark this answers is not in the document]
ccoVeille: What do you think @dnephin ?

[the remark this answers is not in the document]
guerinoni: For me it is right this way, because in this scenario (run, not watch mode)
- start a ctx
- create a process
- wait the execution + testjson does something
- at some point wait finishes
- program goes to the end
- ctx is canceled
- process is stop
- finish

To answer to your concerns, the go-routine stops also in the good path (when all goes well) when the program finish and clearly when the program is interrupted

cmd/main.go
```diff
@@ -510,40 +509,28 @@ func (e exitError) ExitCode() int {
...
 type cancelWaiter struct {
-	cancel  func()
 	wrapped waiter
 }
 
 func (w *cancelWaiter) Wait() error {
 	err := w.wrapped.Wait()
-	w.cancel()
 	return err
 }
```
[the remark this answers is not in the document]
guerinoni: You're right, i removed it now

ccoVeille (approved):

guerinoni:
@dnephin if you have other concerns let me know :D

guerinoni:
@ccoVeille do you think i need other changes here?

ccoVeille:
Nope, the PR is fine and could be merged by a maintainer according to me.

Issue #503, opened 2025-06-13
`gotestsum --version` reports dev instead of installed version `v1.12.2`


jongbeom-moloco:
I installed `gotestsum` using the following command:
```go
"go install gotest.tools/gotestsum@v1.12.2
```

However, when I run the version command, it shows dev instead of the expected version `v1.12.2`:

```sh
> gotestsum --version

gotestsum version dev
```

dnephin:
Thanks! The code that reports version is still relying on it being passed at build time, which doesn't happen when you `go install`.

At some point over the last few years Go added support to capture the git version in the binary. `gotestsum` can be updated to use that to report the version.

Tonkpils:
@devleejb can this be closed given your PR has been merged?

PR #520, opened 2025-08-16
Show correct version when installed via `go install`

devleejb:
Currently, the version information is provided via `ldflags`, which does not work for users who install `gotestsum` using `go install`. This PR changes the version retrieval to use the `fmt/debug` package, allowing users to see the correct version information regardless of the installation method.

Example output for a local build:
```bash
> gotestsum --version
gotestsum version (devel)
```

Example output when installed with go install:
```bash
> gotestsum --version
gotestsum version v1.7.5
```

Related Issue: https://github.com/gotestyourself/gotestsum/issues/503

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"runtime/debug"
 	"strings"
 	"sync/atomic"
 	"syscall"
@@ -17,8 +18,6 @@ import (
 	"gotest.tools/gotestsum/internal/log"
 	"gotest.tools/gotestsum/testjson"
 )
-
-var version = "dev"
 
 func Run(name string, args []string) error {
 	flags, opts := setupFlags(name)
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -34,8 +33,12 @@ func Run(name string, args []string) error {
...
 	switch {
 	case opts.version:
-		fmt.Fprintf(os.Stdout, "gotestsum version %s\n", version)
-		return nil
+		info, ok := debug.ReadBuildInfo()
+		if !ok {
+			return fmt.Errorf("failed to read version info")
+		} else {
+			fmt.Fprintf(os.Stdout, "gotestsum version %s\n", info.Main.Version)
+		}
```
ccoVeille, line 41: There is no need for an else statement when you return

```suggestion
		if !ok {
			return fmt.Errorf("failed to read version info")
		}

		fmt.Fprintf(os.Stdout, "gotestsum version %s\n", info.Main.Version)
```

devleejb: Updated. Thank you for the suggestion.

ccoVeille: Thanks for updating the PR.

Bow, let's wait for the feedbacks of a maintainer

dnephin (approved):
Thank you!

jongbeom-moloco:
@dnephin Thanks for the review. May I ask when the release is planned?

Issue #513, opened 2025-07-16
Prepare for new go 1.25 `T.Attr`


noBlubb:
This is a bit early to the party, but I am actually looking really forward to it: Go 1.25 introduces support [^1] for custom attributes with `T.Attr`.

## Setup
Using the `go1.25rc2`, a test in an otherwise empty project like this

```go
package main

import "testing"

func TestName(t *testing.T) {
	t.Attr("abc", "123")
	t.Attr("yxt", "456")
}
```
with `gotestsum --format=standard-verbose ./...` (at 1.12.1) runs as
```
gotestsum --format=standard-verbose ./...
=== RUN   TestName
=== ATTR  TestName abc 123
=== ATTR  TestName yxt 456
--- PASS: TestName (0.00s)
PASS
ok      go125preview    (cached)

DONE 3 tests in 0.202s
```

where it really only ran one test and not three!
The raw output looks like this:
```
go1.25rc2 test -json                
{"Time":"2025-07-16T21:35:46.672357+02:00","Action":"start","Package":"go125preview"}
{"Time":"2025-07-16T21:35:46.951434+02:00","Action":"run","Package":"go125preview","Test":"TestName"}
{"Time":"2025-07-16T21:35:46.951481+02:00","Action":"output","Package":"go125preview","Test":"TestName","Output":"=== RUN   TestName\n"}
{"Time":"2025-07-16T21:35:46.95152+02:00","Action":"attr","Package":"go125preview","Test":"TestName","Key":"abc","Value":"123"}
{"Time":"2025-07-16T21:35:46.951522+02:00","Action":"output","Package":"go125preview","Test":"TestName","Output":"=== ATTR  TestName abc 123\n"}
{"Time":"2025-07-16T21:35:46.951526+02:00","Action":"attr","Package":"go125preview","Test":"TestName","Key":"yxt","Value":"456"}
{"Time":"2025-07-16T21:35:46.951528+02:00","Action":"output","Package":"go125preview","Test":"TestName","Output":"=== ATTR  TestName yxt 456\n"}
{"Time":"2025-07-16T21:35:46.951533+02:00","Action":"output","Package":"go125preview","Test":"TestName","Output":"--- PASS: TestName (0.00s)\n"}
{"Time":"2025-07-16T21:35:46.951535+02:00","Action":"pass","Package":"go125preview","Test":"TestName","Elapsed":0}
[...]

PR #522, opened 2025-08-26
Add support for go test attributes

noBlubb:
https://github.com/gotestyourself/gotestsum/issues/513

- [ ] Summary counts "attr" events as test executions (fall through in (*Package).addTestEvent marks them as ended)
- [ ] Include attributes in JUnit output


internal/junitxml/report.go
```diff
@@ -224,11 +232,26 @@ func packageTestCases(pkg *testjson.Package, formatClassname FormatFunc) []JUnit
...
+
+	if len(tc.Attributes) > 0 {
+		var properties []JUnitProperty
+		for k, v := range tc.Attributes {
+			p := JUnitProperty{Name: k, Value: v}
+			properties = append(properties, p)
+		}
+		sort.Slice(properties, func(i, j int) bool {
+			return properties[i].Name < properties[j].Name
+		})
```
ccoVeille, line 249: sort.Slice is somehow discouraged.

https://pkg.go.dev/sort#Sort
> Note: in many situations, the newer slices.SortFunc function is more ergonomic and runs faster. 

[slices.SortFunc](https://pkg.go.dev/slices#SortFunc) is a good option.

Also in the function you can use cmp.Compare that can be used directly

https://pkg.go.dev/cmp#Compare

So it ends up to something like this (pseudo code written on a phone)

```suggestion
		slices.Sort(properties, func(a, b JUnitProperty) int {
			return cmp.Compare(a.Name, b.Name)
		})
```

noBlubb:
Hey @dnephin, sorry about the ping but I am looking to land this and want make to sure it's on your radar - LMK if there is anything missing :)

PR #528, opened 2025-09-11
Allow version to be configured at build time

alebcay:
After https://github.com/gotestyourself/gotestsum/pull/520, the use case of setting the version when building from a Git tarball (which contains the repo contents, but not the VCS information) is broken, leaving the version as `(devel)`.

This change returns the option to manually set the version via ldflags. If this manual value is not set (remains empty), then the value from `debug.ReadBuildInfo()` is used.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,6 +19,8 @@ import (
 	"gotest.tools/gotestsum/testjson"
 )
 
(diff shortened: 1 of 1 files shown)


dnephin (approved):
Thank you!

Issue #316, opened 2023-03-29
Add github-actions and buildkite formats


dnephin:
Github actions supports [grouping log lines](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines)

Buildkite supports [managing log output](https://buildkite.com/docs/pipelines/managing-log-output)

Both of these are similar features, they allow some log output to be collapsed (hidden), and expanded by clicking on an arrow. `gotestsum` could use these features to provide better output when tests on run on these sytsems.

See https://github.com/gotestyourself/gotestsum/issues/312#issuecomment-1488048968 for some discussion about a buildkite format, and #315 for a first draft of the github actions format.

Some options for these formats:

1. A format similar to the current `testname` format. Each test would be a group, and expanding the group would show the full output of the test. Currently only failed tests have output, but it wouldn't be too difficult to change to all tests having output hidden by the group.
2. A format similar to the current `pkgname` format. Each package would be a group, and expanding the group would show the full list of tests in that package, or maybe only the failed test output.

dnephin:
I haven't been able to find similar features in other CI systems, but if anyone knows of similar features please do leave a comment with a link to the docs!

Pawka:
> Is there anywhere I can see your build output to get a better idea of what sections are collapsed? Or could you describe it in more detail?

Answering comment from the [previous discussion](https://github.com/gotestyourself/gotestsum/issues/312#issuecomment-1488914280) (since it make more sense to keep context here). I think following rules would work:

1. Grouping should be made per root test. E.g. subtest should not be logged as separate groups.
2. Succeeded test gropus are collapsed.
3. Groups with failed tests are expanded.

However there are some edge cases which should be handled such as timeouts or panics which does not make test as "failed". In such cases code block where failure happened also should be expanded.

Tonkpils:
This seems to be done by https://github.com/gotestyourself/gotestsum/pull/315. Can we close this?

phisco:
> * Grouping should be made per root test. E.g. subtest should not be logged as separate groups.

I arrived here looking exactly for this, it wasn't implemented in the end, right? CC @vitaliyf @dnephin

dnephin:
At this time there is no way to group by root test, that is correct. 

If someone is interested, that sounds like a nice addition as either a new format, or an option to modify the existing github actions format.

PR #537, opened 2025-10-31
feat: buildkite output format

laurynaslubys:
This PR introduces two new output formats that properly collapse test output on [buildkite](https://buildkite.com/docs/pipelines/configure/managing-log-output). There's a quitet variant (`--buildkite`) that suppresses output from passed tests and a verbose one (`--buildkite-verbose`) that prints everything.

I also added some plumbing to enable overriding the summary for a specific format. This is needed to put the summary into a separate section in buildkite rather that under the last test output.

Addresses #316 partly


derekbassett:
What is the status of this?

hiporox:
Any updates?

Issue #540, opened 2025-11-11
Add Github Actions Reporter as a format


jasikpark:
Jest has implemented Github Actions reporter as an output format, which when run in a Github Action, allows Github to then surface test errors alongside the code under review.

Many lint tools have this set up as well, it makes surfacing which tests failed and where they failed just a bit easier.

## Refs: 

- <https://jestjs.io/docs/configuration#reporters-arraymodulename--modulename-options>
- https://github.com/jestjs/jest/blob/main/packages/jest-reporters/src/GitHubActionsReporter.ts
- https://jestjs.io/blog/2022/04/25/jest-28#github-actions-reporter
- https://github.com/jestjs/jest/pull/11320
- https://github.com/jestjs/jest/issues/10309
- https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md
- Minimal example: https://github.com/stefanbuck/jest-matcher/pull/2/files

Tonkpils:
@jasikpark I took a stab at implementing this https://github.com/gotestyourself/gotestsum/pull/544 let me know if it works for you 👍

PR #544, opened 2025-11-21
Implement error reporter format for GitHub Actions

Tonkpils:
- https://github.com/gotestyourself/gotestsum/issues/540

Not sure how else to showcase it without having failing CI builds but here's an output example

image

This pull request adds support for emitting GitHub Actions error annotations for failed tests in the `githubActionsFormat` output. The main changes include parsing test output for failure locations, handling panics specially, and generating error annotations with file and line information for easier debugging in CI environments.

**GitHub Actions error annotation support:**

* Added regex-based parsing of test output to extract file and line information for failed tests and panics (`fileLinePattern`, `panicStackPattern` in `testjson/format.go`). [[1]](diffhunk://#diff-26fa7a88f4cd9bae3f9b2abbe8b3a72a1f337411859d3fa8d1439670b577be1cR8-R9) [[2]](diffhunk://#diff-26fa7a88f4cd9bae3f9b2abbe8b3a72a1f337411859d3fa8d1439670b577be1cR468-R473)
* Introduced the `writeGitHubActionsError` function to emit error annotations in GitHub Actions format, including special handling for panics and regular failures, with proper sanitization of annotation fields (`testjson/format.go`).
* Modified the test formatter to call `writeGitHubActionsError` on failed test events, resulting in error annotations for each failure (`testjson/format.go`).

**Test output updates:**

* Updated expected output in `testjson/testdata/format/github-actions.out` to include new error annotations for each failed test, showing file, line, and test name information. [[1]](diffhunk://#diff-e718a0b18632e33878e803c8a1f8bf2939680e7c18c977af1966f956e99569f1R77-R113) [[2]](diffhunk://#diff-e718a0b18632e33878e803c8a1f8bf2939680e7c18c977af1966f956e99569f1R137) [[3]](diffhunk://#diff-e718a0b18632e33878e803c8a1f8bf2939680e7c18c977af1966f956e99569f1R146) [[4]](diffhunk://#diff-e718a0b18632e33878e803c8a1f8bf2939680e7c18c977af1966f956e99569f1R168) [[5]](diffhunk://#diff-e718a0b18632e33878e803c8a1f8bf2939680e7c18c977af1966f956e99569f1R182) [[6]](diffhunk://#diff-e718a0b18632e33878e803c8a1f8bf2939680e7c18c977af1966f956e99569f1L213)

### DIFF
--- a/testjson/format.go
+++ b/testjson/format.go
@@ -602,7 +602,7 @@ func writeGitHubActionsError(
 		}
 	} else {
 		// For regular test failures, emit one annotation per error line
-		for _, outputLine := range outputLines {
+		for idx, outputLine := range outputLines {
 			if matches := patterns.fileLine.FindStringSubmatch(outputLine); len(matches) == 3 {
 				file := matches[1]
 				line := matches[2]
@@ -620,6 +620,9 @@ func writeGitHubActionsError(
 					message = strings.TrimSpace(parts[2])
 				}
 				if message == "" {
+					message = collectAdditionalMessage(outputLines[idx+1:], patterns)
+				}
+				if message == "" {
 					message = "Test failed"
 				}
 
@@ -628,4 +631,32 @@ func writeGitHubActionsError(
 			}
 		}
 	}
+}
+
+func collectAdditionalMessage(lines []string, patterns githubActionsErrorPatterns) string {
+	var parts []string
+	shouldStop := func(line string, trimmed string) bool {
+		if trimmed == "" {
+			return true
+		}
+		if patterns.fileLine.MatchString(line) || patterns.panicStack.MatchString(line) {
+			return true
+		}
+		if strings.HasPrefix(trimmed, "PASS ") || strings.HasPrefix(trimmed, "FAIL ") ||
+			strings.HasPrefix(trimmed, "SKIP ") || strings.HasPrefix(trimmed, "=== ") ||
+			strings.HasPrefix(trimmed, "--- ") || strings.HasPrefix(trimmed, "::") {
+			return true
+		}
+		return false
+	}
+
+	for _, line := range lines {
+		trimmed := strings.TrimSpace(line)
+		if shouldStop(line, trimmed) {
+			break
+		}
+		parts = append(parts, trimmed)
+	}
+
+	return strings.Join(parts, " ")
 }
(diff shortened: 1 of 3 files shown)


testjson/format.go
```diff
@@ -513,3 +526,85 @@ func githubActionsFormat(out io.Writer) EventFormatter {
...
+		s = strings.ReplaceAll(s, "\r", "%0D")
+		s = strings.ReplaceAll(s, "\n", "%0A")
+		return s
+	}
+
+	// Check if this is a panic by looking for panic: in the output
+	var isPanic bool
+	var panicMessage strings.Builder
+	for _, outputLine := range outputLines {
+		if strings.Contains(outputLine, "panic:") {
```
ccoVeille: I'm a bit worried this would be true for anything returning the string `panic:`

The next loop and if would use the regexp to extract the info

Maybe you could check for "panic:", loop, then test with regexp, and you match once, apply the panic/test results

Tonkpils: Do you mean to do this process

> "panic:", loop, then test with regexp, and you match once, apply the panic/test results


For every time we see the `panic:` match or only the once?

What this does is to gather up all the messages that match `panic:` then attributes them to the matching stack trace in the next if/loop block. 

Happy to make the necessary change. Just want to make sure I'm writing it up for the correct scenario you're thinking about 😅

ccoVeille: I'm not sure either what I meant 😬 

No matter how it could implemented I just want that gotestsum doesn't report a panic if someone called

```go
t.Errorf("checking there is no panic: just in case")
```

I feel like adding a test about it would help you to figure how to catch the issue I thought about

Tonkpils: That makes sense. I can make the panic check a bit more strict but unfortunately I don't believe there's much we can do to prevent someone from writing an error message that may be similar to a panic initial error message.

I can change the condition to check if the entire output line starts out with `panic:` which is more strict than checking for a substring of `panic:` in the entire message. 

Does that sound reasonable? 

Either way, I'll see if I can add some more specific tests to this format so we can have a better picture of the setup/expectation.

ccoVeille: Let's wait for the feedback of maintainers here.

I feel like I'm the only one who reviewed your PR right now

ccoVeille (approved):

Tonkpils:
@dnephin 👋 sorry for the direct ping, just wanted to make sure you didn't miss this PR. Do you have any feedback on whether this is good to be merged or if you'd like to see any changes? 🙇

PR #558, opened 2026-02-20
New formatter `github-actions-concise`

kalverra:
Adds a `github-actions-concise` formatter that allows for a much smaller footprint when viewing test results in CI. You often want to see only the failed tests in CI, and having to scroll past hundreds of `PASS` messages in a large suite isn't great UX.

[See an example run using the old and new formats](https://github.com/kalverra/gotestsum/actions/runs/22231705882/job/64312828842)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,12 +7,7 @@ on:
 
 jobs:
   Build:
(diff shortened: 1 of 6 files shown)


kalverra:
Closing PR as stale

PR #560, opened 2026-03-23
Add env var support for --format-hide-empty-pkg

endigma:
## Summary

- Add `GOTESTSUM_FORMAT_HIDE_EMPTY_PKG` environment variable to configure the `--format-hide-empty-pkg` flag
- Uses the existing `truthyFlag(lookEnvWithDefault(...))` pattern already used by `--junitfile-hide-empty-pkg` and `--junitfile-hide-skipped-tests`
- Accepts `true`, `yes`, or `1` (case-insensitive); defaults to `false` when unset

## Motivation

The `--format-hide-empty-pkg` flag is useful for coding LLMs reading gotestsum output. The easiest way to force LLM run gotestsum invocations to use this option is an environment variable.


endigma:
Hi @dnephin, anything I can do to help move this along?

PR #563, opened 2026-04-10
Remove unused Go version checks and related test data

alexandear:
All these files are unused since the tool supports [Go 1.24+](https://github.com/gotestyourself/gotestsum/blob/06f60b3249917ddd1a04c8e2586116d1a87bc67c/go.mod#L3).


dnephin (approved):
Thank you!

PR #566, opened 2026-04-27
rerun-fails: detect parent-test runtime.Goexit as incomplete run

### LINKED ISSUE #565: gotestsum --rerun-fails can exit successfully after incomplete parent test run
## Summary

`--rerun-fails` can report success after Go aborts part of a subtest tree with a parent-test `runtime.Goexit`.

This happens when a subtest calls `FailNow`/`Fatal` on a parent `*testing.T`. Go emits:

`test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test`

At that point, sibling subtests may never run. gotestsum then reruns only the failed tests it observed. If those reruns pass, gotestsum exits 0 even though the original selected test tree did not fully execute.

## Why This Is Unsafe

This is similar to the existing suspected-panic guard: the first run is incomplete, so the set of failed tests is not a safe rerun set.

The final output can look like:

```text
DONE 2 runs, 9 tests, 3 failures in 213.360s
```

but the process still exits 0 because the rerun passed.

## Expected Behavior

When gotestsum sees the parent-test `runtime.Goexit` diagnostic, `--rerun-fails` should abort with a non-zero exit rather than rerunning only observed failures.

## Related

- https://github.com/gotestyourself/gotestsum/issues/290
- https://github.com/golang/go/issues/45508

chandler-solo:
## Summary

Treat Go's parent-test `runtime.Goexit` diagnostic as an incomplete test run when using `--rerun-fails`.

If a subtest calls `FailNow`/`Fatal` on a parent `*testing.T`, Go emits:

`test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test`

That can abort the rest of the test branch, so some tests may never run. In that state, rerunning only the observed failed tests is unsafe because gotestsum can report success after the rerun passes while skipped sibling tests were never executed.

Fixes #565 

## Changes

- Detect the `runtime.Goexit` diagnostic alongside existing suspected panic detection.
- Abort reruns with the existing incomplete-run guard.
- Add a regression test covering the parent-test `runtime.Goexit` case.

## Testing

```sh
go test ./cmd ./testjson
go test ./...
```

### DIFF
--- a/cmd/main_test.go
+++ b/cmd/main_test.go
@@ -441,6 +441,44 @@ func TestRun_RerunFails_PanicPreventsRerun(t *testing.T) {
 	assert.ErrorContains(t, err, "rerun aborted because previous run had a suspected panic", out.String())
 }
 
+func TestRun_RerunFails_RuntimeGoexitPreventsRerun(t *testing.T) {
+	jsonFailed := `{"Package": "pkg", "Action": "run"}
+{"Package": "pkg", "Test": "TestOne", "Action": "run"}
+{"Package": "pkg", "Test": "TestOne/Subtest", "Action": "run"}
+{"Package": "pkg", "Test": "TestOne/Subtest", "Action": "output","Output":"    testing.go:1913: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test\n"}
+{"Package": "pkg", "Test": "TestOne/Subtest", "Action": "fail"}
+{"Package": "pkg", "Test": "TestOne", "Action": "fail"}
+{"Package": "pkg", "Action": "fail"}
+`
+
+	var calls int
+	fn := func([]string) *proc {
+		calls++
+		return &proc{
+			cmd:    fakeWaiter{result: newExitCode("failed", 1)},
+			stdout: strings.NewReader(jsonFailed),
+			stderr: bytes.NewReader(nil),
+		}
+	}
+	reset := patchStartGoTestFn(fn)
+	defer reset()
+
+	out := new(bytes.Buffer)
+	opts := &options{
+		rawCommand:                   true,
+		args:                         []string{"./test.test"},
+		format:                       "testname",
+		rerunFailsMaxAttempts:        3,
+		rerunFailsMaxInitialFailures: 1,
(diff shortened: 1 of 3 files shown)


MikeeI:
The fix blocks the real parent-`Goexit` case end to end, but the unanchored fragment match also aborts valid reruns when ordinary test output contains that text.

I reproduced this on Go 1.26.5 with the PR binary. A test that logs the fragment, fails once via a marker file, and passes on rerun behaves as follows:

```go
func TestLoggedText(t *testing.T) {
    t.Log("test executed panic(nil) or runtime.Goexit")
    if _, err := os.Stat("first-run"); err == nil {
        return
    }
    if err := os.WriteFile("first-run", nil, 0o600); err != nil {
        t.Fatal(err)
    }
    t.Fail()
}
```

With gotestsum v1.13.0, `--rerun-fails=1` reruns the test and exits 0. With this PR, the first run is classified as incomplete and gotestsum exits 3 without rerunning:

```text
ERROR rerun aborted because previous run had a suspected panic or runtime.Goexit and some test may not have run
exit=3
```

The real Go 1.24–1.26 diagnostic includes the complete suffix:

```text
test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

and is emitted from `testing.go:<line>`. Matching the complete diagnostic line, including that suffix and its `testing.go:` source prefix while allowing the line number to vary, would avoid treating arbitrary user output as an incomplete run.

I also ran the original parent-`Goexit` scenario against the PR binary: it exits 3 before a rerun and the sibling-test sentinel remains absent, so the intended guard works.

### Disclosure

Investigated thoroughly with openai-codex/gpt-5.6-sol (medium reasoning effort), using [Oh My Pi](https://github.com/can1357/oh-my-pi) as the agent framework.

This report is not generic or unreviewed AI-generated output. Its claims were checked against the cited evidence, and it includes the relevant detail intended to help maintainers resolve the issue.

If reports like this are not useful to the project, please let me know and I will refrain from submitting similar ones. My intent is to help without wasting maintainer time or energy or discouraging their work.

Thank you for your work.

PR #567, opened 2026-05-15
Fix compilation for AIX

pgimalac:
The `Event` struct in `watch_unsupported.go` (AIX build stub) was missing the `Args []string` field that was added to the main `Event` struct.

This causes a compilation failure on AIX:
```
cmd/watch.go: event.Args undefined (type filewatcher.Event has no field or method Args)
```

This PR fixes the build on AIX/ppc64.

### DIFF
--- a/internal/filewatcher/watch_unsupported.go
+++ b/internal/filewatcher/watch_unsupported.go
@@ -11,6 +11,7 @@ import (
 
 type Event struct {
 	PkgPath string
+	Args    []string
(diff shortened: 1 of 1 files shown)


pgimalac:
The `Event` struct could just be moved to a common file by the way, it doesn't contain anything OS specific.

dloucasfx:
bump
Running into the same issue and this addresses the issue

PR #568, opened 2026-05-23
feat: Add GitHub Actions failure formatter

### LINKED ISSUE #543: Allow GitHub actions formatter to only display failure outputs.
Using the GitHub Actions formatter currently logs out the output of every test. This is may be undesirable if there's a lot of output since Actions will truncate it and require downloading the raw log. 

Instead, it'd be ideal to allow configuring it or having a separate formatter to only display the logs on failure tests.

abhinav:
Large test suites can overload GitHub Actions logs,
slowing down log browsing in the UI.

Add `github-actions-fails` as a quieter formatter
that keeps package progress visible
while only expanding failed test diagnostics.

Package-level failures still promote buffered package output so
TestMain and init failures keep their error context.

Naming: this is called github-actions-fails to mirror
the precedent set by pkgname-and-test-fails.

Resolves #543


abhinav:
This change is part of the following stack:

- #568 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

abhinav:
I see there's also https://github.com/gotestyourself/gotestsum/pull/558 implementing a similar feature.

